PR #10, opened 2016-01-26
Update CMakeLists.txt

gsjaardema:
Target link libraries need to be set on systems other than just WIN32 and CYGWIN.  

Not sure why preview is showing that everything changed; I only changed code in the line range 474 to 486.


gsjaardema:
This change seems to be required on MacOSX, but linux seems to work with or without it.

brtnfld:
Looks like an issue with with space, the change is:

474d473
<   endif(WIND32 OR CYGWIN)
486a486
>   endif (WIN32 OR CYGWIN)

brtnfld:
I went ahead and just made the changes by hand.

gsjaardema:
Thanks, yes there were a couple typos in that line; 
Thanks

PR #11, opened 2016-01-26
Support Mac's dylib shared library suffix

gsjaardema:
Use a predefined CMake variable to specify the shared library suffix (.so vs .dylib) to facilitate building on MacOSX


brtnfld:
Thanks, I've added the enable build shared option for daily testing.

PR #14, opened 2016-08-15
Add missing include path

gsjaardema:
CMake-based build of tkogl was missing include path for opengl headers.  Not sure this is the correct fix, but it eliminates this and several other errors about not being able to find GL/gl.h

### DIFF
--- a/src/cgnstools/tkogl/CMakeLists.txt
+++ b/src/cgnstools/tkogl/CMakeLists.txt
@@ -2,7 +2,7 @@
 # tkogl #
 #########
 
-include_directories(${TCL_INCLUDE_PATH} ${TK_INCLUDE_PATH})
+include_directories(${TCL_INCLUDE_PATH} ${TK_INCLUDE_PATH} ${OPENGL_INCLUDE_DIR})
 
 set(tkogl_FILES 
 	tkogl.c


brtnfld:
I'm wondering if should redo the include, from stackoverflow:

http://stackoverflow.com/questions/16192775/cmake-opengl-include-dirs-are-named-differently-on-different-platforms

You have a couple of choices here. The simplest is to add ${OPENGL_INCLUDE_DIR}/GL or ${OPENGL_INCLUDE_DIR}/OpenGL to your include search paths and use

#include "gl.h"

in your source code.

The case of filenames is disregarded on Windows, so you don't need to use /gl and /GL - simply /GL will do.

To achieve this, in your CMakeLists.txt do:

if(APPLE)
  include_directories(${OPENGL_INCLUDE_DIR}/OpenGL)
else()
  include_directories(${OPENGL_INCLUDE_DIR}/GL)
endif()


To make this a bit more robust, you can have CMake find the path to "gl.h" and include that:

find_path(OpenglIncludeSubdir
            NAMES gl.h
            PATHS ${OPENGL_INCLUDE_DIR}
            PATH_SUFFIXES GL OpenGL
            NO_DEFAULT_PATH)
include_directories(${OpenglIncludeSubdir})

Again, in your source you'd use:

#include "gl.h"


Possibly the most robust (i.e. avoiding further include paths beyond ${OPENGL_INCLUDE_DIR}) would be to find the path to "gl.h" and then use configure_file to apply the correct #include term. You would use an input file which contained something like:

#include "@OpenglSubdir@/gl.h"

and then running configure_file would replace @OpenglSubdir@ with its value and write the contents to an output file. This output file would then be included in your target and would have

#include "GL/gl.h"

or

#include "OpenGL/gl.h"

as appropriate.

To achieve this, you'd do something like:

find_file(OpenglSubdir
            NAMES GL OpenGL
            PATHS ${OPENGL_INCLUDE_DIR}
            NO_DEFAULT_PATH)
get_filename_component(OpenglSubdir ${OpenglSubdir} NAME)
configure_file(${CMAKE_SOURCE_DIR}/my_config.h.in ${CMAKE_SOURCE_DIR}/my_config.h)

gsjaardema:
Yes, something like that would probably work better than my change.  I like the "... a bit more robust..." solution near the middle, but any of the options look like they would work.

brtnfld:
Added task to redo the include
https://cgnsorg.atlassian.net/browse/CGNS-106

gsjaardema:
Any update on this?  I still need to manually make this change whenver I want to build the tools on my Mac...

brtnfld:
Sorry, this dropped off my radar, I'll try to look into it next week.

brtnfld:
We are addressing this in PR #181

PR #18, opened 2016-08-16
Add missing include path, fix hdf5

gsjaardema:
Added missing opengl_include_dir path to include_directories.

With the current hdf5 find_package, it doesn't seem like HDF5_LIBRARY is defined and the link dependency is specified earlier on the CGNS library itself, so removed HDF5_LIBRARY from the test and the link library.  

With these changes (and other pull requests), this builds in both shared and static mode on a Mac.


brtnfld:
May need to 'fix' the OPENGL_INCLUDE_DIR option to be more robust.

PR #29, opened 2016-12-21
CGNS-85 fix,

jm-cc:
cgp_open now uses current communicator, defined by cgp_mpi_comm.

Hi, I ran into CGNS-85 also in the case of a MPMD program where rank 0 in MPI_COMM_WORLD was not involved in opening the file.
Here is the fix I propose, similar to the one described in the CGNS-85 issue. 
Basically, the communicator used when calling cgp_mpi_comm is saved and reused for cgp_open and cgio_check_file.

### DIFF
--- a/src/cgns_io.c
+++ b/src/cgns_io.c
@@ -47,6 +47,7 @@ char hdf5_access[64] = "NATIVE";
 #endif
 #ifdef BUILD_PARALLEL
 #include <mpi.h>
+MPI_Comm pcg_mpi_comm=MPI_COMM_WORLD;
 int pcg_mpi_comm_size;
 int pcg_mpi_comm_rank;
 int pcg_mpi_initialized;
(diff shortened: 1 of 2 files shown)


brtnfld:
I'll work on a test for the patch this weekend if you already don't have one.

jm-cc:
Hi, I've got a very basic test that deadlocks without the fix. I am not sure it is good enough to be included, I just did it to mimic my app behaviour.
First it creates a file with every processes. Then it creates a communicator that does not contain Process 0 of MPI_COMM_WORLD and try to access the file.
```C
#include <stdio.h>
#include <string.h>
#include "pcgnslib.h"
#include <mpi.h>

int main(int argc, char** argv)
{
  MPI_Comm dummy;
  char test[20];
  int index_file;
  int rank; //rank in MPI_COMM_WORLD
    
  snprintf(test, 20, "test.cgns");

  MPI_Init(&argc, &argv);

  cgp_open(test, CG_MODE_MODIFY, &index_file);
  cgp_close(index_file);


  MPI_Comm_rank(MPI_COMM_WORLD, &rank);
  if (rank==0) 
    MPI_Comm_split(MPI_COMM_WORLD, 0, 0, &dummy);
  else
    MPI_Comm_split(MPI_COMM_WORLD, 1, rank, &dummy);

  if (rank>0) //If member of the dummy communicator
  {
    printf("%d\n",rank);
    cgp_mpi_comm(dummy);
    cgp_open(test, CG_MODE_MODIFY, &index_file);
    cgp_close(index_file);
  }
  MPI_Finalize();
}
```

PR #30, opened 2017-01-06
Spelling error in function name

akilan27:
This corrects spelling mistake in `HashStats()` function in two different files. I discovered this while trying to use the function but got linker errors.

PS: This same file is present in **two** different directories. Code should be modified to use one of those & delete the other file. This will help code maintenance. I can send another pull request for that change if that change is desired.


brtnfld:
Changes are unrelated to travis CI failures, merging.

PR #33, opened 2017-01-17
Prototype for cgp_parent_data_write

gsjaardema:
A first shot at an implementation for cgp_parent_data_write. It works in the testing I have done on my local project, but not sure if the "style" is correct. It currently uses an MPI_Allreduce call to determine the total size of the ParentElements and ParentElementsPosition nodes which are created during the function. It might be cleaner to have two functions -- one to create the nodes and one to write the data to the nodes, but this implementation follows the behavior of the serial version.

Let me know what you think and I will modify as needed.

### DIFF
--- a/src/pcgnslib.c
+++ b/src/pcgnslib.c
@@ -508,8 +508,7 @@ int cgp_elements_read_data(int fn, int B, int Z, int S, cgsize_t start,
 }
 
 int cgp_parent_data_write(int fn, int B, int Z, int S,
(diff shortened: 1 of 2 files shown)


brtnfld:
Can you add a test for the new API?

gsjaardema:
Yes, I will take a look and hopefully be able to add a test in the next
week or so.
..Greg

gsjaardema:
This has been moved to #36 which has both the implementation and a test.

PR #38, opened 2017-02-28
Add option to permit experimenting with HDF5 "DISKLESS" mode

gsjaardema:
We have a couple projects in which we are experimenting with using the "DISKLESS" hdf5 mode.  This is described at https://support.hdfgroup.org/HDF5/doc/RM/H5P/H5Pset_fapl_core.htm.  The addition of the three lines in this pull request lets us test the viability of this capability with minimal changes to the CGNS library.

An application would specify this mode by adding `extern char hdf5_access[64];` to the application and then doing something like:

```
int old_type = cgns_filetype;
strcpy(hdf5_access,"DISKLESS");    
int ierr = cg_set_file_type(CG_FILE_HDF5);
ierr = cg_open(get_filename().c_str(), mode, &cgnsFilePtr);
strcpy(hdf5_access,"NATIVE");
cgns_filetype = old_type;
```

This mirrors somewhat the code in `cgp_open()` 

If this looks like a capability that would be generally useful to other users of the library, then this capability could be added to a new API function, but for now the minimal change here would be all that is needed.  I understand that there may be some hesitation at putting this in; if so, then we can probably handle it by creating an external copy of the ADFH_Database_Open function, but that would be much more work...

Thanks for any consideration or feedback on this suggested change; perhaps there is a better way to accomplish what we are trying to do.


brtnfld:
gcc -g -g -O2   -I.  -DBUILD_HDF5 -Iadfh -I/home/travis/hdf5/include  -o lib/ADFH.o -c adfh/ADFH.c

adfh/ADFH.c: In function ‘ADFH_Database_Open’:

adfh/ADFH.c:2153:44: error: ‘true’ undeclared (first use in this function)

adfh/ADFH.c:2153:44: note: each undeclared identifier is reported only once for each function it appears in

make: *** [lib/ADFH.o] Error 1

brtnfld:
I've added the DISKLESS option to CGNS-220 to investigate creating all the CGNS metadata DISKLESS, and then opening the file in parallel for the raw writes.

brtnfld:
Created issue to track:

CGNS-239
Add diskless option for creating a CGNS file

PR #39, opened 2017-03-01
Add cmake support for memory debugging; fix memory debug routines

gsjaardema:
This PR permits the control of the memory debugging capability via cmake.  To enable, define CGNS_ENABLE_MEM_DEBUG=ON.

The address storing in cg_malloc did not seem to work correctly; perhaps an issue with sequence points, so separated the `*data++ = (size_t)(data + 1);` into two lines: `*data = (size_t)(data + 1);` and `data++;`

Slight modification to the memory debug output:
* Added "CGNS MEM_DEBUG" prefix to the output lines in cg_open and cg_close.
* Added output of the number of alloc and free calls.


gsjaardema:
Thanks,

PR #45, opened 2017-08-25
_stat32i64 is msvc-specific

jschueller:


mennodeij:
It was already inside a #ifdef _WIN32 statement, but I guess that is too general? 
I added the #define stat _stat32i64 to enable the loading of large files on Windows. Make sure to test that for your situation.

jschueller:
yeah, but that's only relevant to the msvc compiler, not for mingw

jschueller:
is it ok to merge ? or ... ?

mennodeij:
As far as I'm concerned yes.

jschueller:
cc @brtnfld

PR #46, opened 2017-10-04
Specify STATIC when adding cgns_static lib

jschueller:
Or else it is sensitive to BUILD_SHARED_LIBS cmake var


jschueller:
ping

jschueller:
ping @brtnfld

PR #49, opened 2017-12-27
Shared lib (dll) can be built on windows. Fixes CGNS-129.

matmel:
The main change is in root CMakeList.txt. I had to change a the location of the if block that adds the HDF5 definitions `-DH5_BUILT_AS_DYNAMIC_LIB` and `-DH5_BUILT_AS_STATIC_LIB`. They are required to properly define HDF5 headers on windows, especially to export/import symbols correctly.

I found this [blog post](http://blog.omega-prime.co.uk/2011/07/04/everything-you-never-wanted-to-know-about-dlls/) useful to understand correctly the dll symbol export mechanism.

Also, I changed the appveyor script to build the shared version of CGNSlib.

### Question 1
In ADFH.h the line 
```
#if defined(WIN32)
```
Should probably be replaced with 
```
#if defined(_MSC_VER)
```
as it is done in HDF5 library (see `H5api_adpt.h` for instance) because the `__declspec(dllexport)` decorator thing is only relevant for MSVC as far as I know. `gcc` on windows does not have this requirement.

### Question 2

Should we update the version of HDF5 (1.10.*) and AppVeyor image to use the latest visual studio?


brtnfld:
I checked your fork's fix with hdf5 develop (i.e. 1.11), which was build with both static and shared HDF5 libraries, and  CMake still fails with linking errors caused by pre-appending __imp_ in front of the HDF5 functions,

For example, 
  cgns.lib(ADFH.obj) : : error LNK2019: unresolved external symbol __imp_H5get_libversion referenced in function ADFH_Library_Version

This is on windows 10, with the Intel C compiler.

matmel:
I installed Intel compilers and tried to compile (and install) HDF5-1.10.1 and cgnslib with them. I had the linking errors that you mentioned with `cmake -G "Visual Studio 15 2017" -T "Intel C++ Compiler 18.0" ...` for `HDF5` and `cmake -G "Visual Studio 15 2017 Win64" -T "Intel C++ Compiler 18.0" ...` (Note the `Win64` in the cmake generator for cgnslib and not hdf5).

When I switched the hdf5 generator to `Visual Studio 15 2017 Win64`, the error disappeared.

mathstuf:
Note that, yes, GCC on Windows does use `__declspec`. Cygwin doesn't, but Cygwin is also not Windows (it just happens to run on top of it).

matmel:
According to this page: https://cmake.org/Wiki/CMake_Checking_Platform, cygwin defines the `UNIX` directive (if the cygwin cmake is used) and not the `WIN32` one. So with this PR, on cygwin, the `__decspec` decorator will not be used and AFAICT this is the right thing to do. I did not test it though, so there may be more to do to make sure cygwin does not fail.

For intel compilers, according to that page: https://software.intel.com/en-us/articles/exporting-using-_declspecdllexport-on-windows, intel C compiler follows the Visual C++ way with respect to `__declspec`, and as I checked yesterday, it works.

I did not test what happens if HDF5 is compiled with visual c++ and cgnslib with intel or vice versa.

brtnfld:
Are you testing with cygwin? I think the issue is we are not currently testing CGNS with cygwin; not with CI or in the daily testing we run. I'm not even sure the cygwin build still works.

matmel:
No I don't test with cygwin and have no particular intention to test it neither :wink:

I just tested with visual c++ and intel c++ and I was not able to reproduce the linking errors with intel except when I unfortunately mixed a 32bit build of HDF5 with a 64bit build of cgnslib.

brtnfld:
looks fine, but we should add multiple testing configurations to appveyor for testing both static and shared builds instead of just doing shared builds.

CMakeLists.txt
```diff
@@ -233,17 +225,29 @@ if (CGNS_ENABLE_HDF5)
...
+      message( STATUS "Adding pre-processor definition H5_BUILT_AS_STATIC_LIB")
+      add_definitions (-DH5_BUILT_AS_STATIC_LIB)
+    endif (HDF5_BUILD_SHARED_LIBS)
+    if (BUILD_SHARED_LIBS AND WIN32)
+      set_property (TARGET h5dump PROPERTY IMPORTED_LOCATION "${HDF5_TOOLS_DIR}/h5dumpdll")
+    else (BUILD_SHARED_LIBS AND WIN32)
+      set_property (TARGET h5dump PROPERTY IMPORTED_LOCATION "${HDF5_TOOLS_DIR}/h5dump")
+    endif (BUILD_SHARED_LIBS AND WIN32)
   else (HDF5_FOUND)
+    message( STATUS 'Hello there D')
```
brtnfld: remove this

src/adfh/ADFH.h
```diff
@@ -174,7 +174,7 @@
 	Prototypes for Interface Routines
 ***********************************************************************/
 
-#if defined(_WIN32) && defined(BUILD_DLL)
+#if defined(WIN32) && defined(BUILD_DLL)
```
brtnfld: I think _WIN32 is correct. Why use WIN32?

mathstuf: Indeed, `_WIN32` is correct. CMake sets `WIN32` for legacy reasons, but the real one should be preferred.

matmel:
Thank you @brtnfld for merging this.

I realized yesterday that I actually build a shared version of the library alongside with the static version. The tests are linked with the static version only. When the dynamic version is built, the tests should link with the dynamic one.

I will open a new PR with your code review comments, the fix in the testing and a build matrix in the appveyor configuration. BTW do you want me to test two versions of the HDF5 library?

brtnfld:
That would be great if you can update appveyor with a matrix build for windows, we could add testing for 1.10.1 if we had a place to download the pre-built binaries from, I don't want to build HDF5 from source.

brtnfld:
The CI develop branch fails with this commit on windows.

PR #51, opened 2018-01-24
Workaround OSX compiler bug

gsjaardema:
The changed code above is functionally equivalent to the original code, but I get HDF5 errors in H5Dwrite with the original code, but not with the changed code on OSX.  This is with clang-3.9.1 specifically, but I also saw with other compilers.  Not sure what the issue is.  I can also get correct behavior by adding a "printf(stderr, "%d %d\n", start, end)" after line 608.  Obviously something strange is happening and I am OK if the PR is rejected since it really adds no true value...  

Just proposing in case others are seeing a similar problem and this can provide a solution.


gsjaardema:
This does not seem to be needed.  #50 seems to have fixed the issue without this PR.

PR #60, opened 2018-04-03
Now compression for HDF5 files is working.

kskinoue0612:
Previously, `cg_configure()` with `CG_CONFIG_HDF5_COMPRESS` option did not work, and this commit will fix that.

I know this implementation about how to decide the size of chunk is not perfect, but acceptable, I hope.

Thanks for developing the CGNS library. It is quite a useful library, and I depend on this in my work.

Thanks,

Keisuke Inoue

### DIFF
--- a/src/adfh/ADFH.c
+++ b/src/adfh/ADFH.c
@@ -2685,7 +2685,7 @@ void ADFH_Put_Dimension_Information(const double   id,
                                     int           *err)
 {
   hid_t hid;
-  hid_t did, tid, sid, mid;
+  hid_t did, tid, sid, mid, pid;
   int i, swap = 0;
   hsize_t old_size;
   hsize_t old_dims[ADF_MAX_DIMENSIONS];
@@ -2693,6 +2693,7 @@ void ADFH_Put_Dimension_Information(const double   id,
   void *data = NULL;
   char new_type[3];
   hid_t xfer_prp = H5P_DEFAULT;
+  hsize_t chunk_dim[ADF_MAX_DIMENSIONS];
 
   to_HDF_ID(id,hid);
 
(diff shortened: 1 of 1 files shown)


kskinoue0612:
Hi,

I've found an issue on this pull request, so please do not merge this now. I'll fix, and contact again.

Thanks,

Keisuke Inoue

kskinoue0612:
Hi,

I'm sorry, it was my misunderstanding that this pull request has an issue.

I thought I tested with dimensions not multiple of 5, and it failed.

But that was not true. This pull request worked with dimensions like 111.

brtnfld:
Unfortunately, the issue with compression is getting chunking correct, which is the main reason why it's not implemented. Picking a single chunk size is not the best solution, you can get situations that the chunks don't align with dataset bounds, so you may cause more I/O than desired. You also have to consider the size of the chunk cache. Chunk dimension sizes should align with possible hyperslab dimensions for I/O. Setting the chunk size by this method will create chunks that are most likely too large. Setting chunk size is mostly about performance, and not necessarily just about getting compression to work (which is trivial once chunking works). Is a given chunk size for an unstructured mesh good for a structured mesh?

You also need to take into account the parallel file system parameters, for example, your chunk size in relation to the stripe size on a Lustre system. Also in parallel, there is a 2GB chunk size limit.

Adding compression/chunking is probably beyond a simple pull request because a lot of different use cases have to be handled. 

For this feature to be implemented, it will probably need a CGNS Proposals for Extension (CPEX) or an RFC, so that we can get feedback from other users and the committee, see http://cgns.github.io/Proposals.html

The proposal would need to address use cases, benchmarks, how to introduce custom compression filters, how does the CGNS library handle a CGNS file with unknown compression filters, how to have a user set chunk size, etc... We also need to have test code. 

If you would like to further discuss this, we can invite you to a CGNS steering committee meeting, and we can decide where to go from there. It would be great to have compression enabled in CGNS, but I think it's more involved than this solution.

kskinoue0612:
Hi Scot,

Thank you for taking your time to comment to my pull request. I've understood the technical detail, and understood that my PR was not good enough to be merged.

I actually tested my PR code with just a few test cases. In those cases my code worked fine and CGNS file size was decreased to be less than 10% of the original size. That was the reason that I made the PR.

Anyway, I understand that chunk size is important.

Currently, chunking related implementation is in ADFH_Put_Dimension_Information, and this function is called every time when an array is written into CGNS file. It will be difficult to guess the good chunk size for every calls for this function, like writing Zone_t, Unstructured Grid coordinates, Structured Grid coordinates, or Solution values. One possibility will be to use different implementation of guessing chunk size for different use cases.

I understand the importance of parallel I/O too. Recently I've worked on ADIOS (https://www.olcf.ornl.gov/center-projects/adios/). Working on this project, I've got to know that recently in HPC region, the low speed of storage is dragging down the speed of whole system. Parallel I/O will be one of the solutions for the issue.

Thanks for your comment to invite me to CGNS steering committee. It is an great honor for me. I live in Japan, and it will be hard for me to join the committee, actually. But, I'm sure I'll go on to be a heavy user of CGNS library, and if I find out something that I can contribute to make CGNS better, I'll try to contribute maybe through making a new proposal to the issue tracking system, or making a pull request in github.

Thanks again,

Keisuke Inoue

PR #61, opened 2018-04-24
Proof-of-concept for allowing collective and independent HDF5 io

gsjaardema:
With the current library, if the parallel API is used, then it is not possible to use the "independent" API in the same execution.  We have a need to support both "collective" in which all mpi ranks are outputting to a single file and "independent" in which each mpi rank is outputting to its own file (file-per-processor).  

As best I can tell, the changes in this PR allow this behavior.  It is currently more of a proof-of-concept than it is production-ready code.  The basic theory is that a `mpi_initialized` member is added to the `cgns_io` struct.

At file creation/open, the state of the global variable `pcg_mpi_initialized` is checked and stored in the cgns_io struct assigned to that file.  Then at each read/write call, the `pcg_mpi_initialized` global variable is set to the value in `cgio->mpi_initialized`

A klugy portion of this is that when the client calls `cg_open` after having previously called `cgp_open`, the client needs to manually set the value of `pcg_mpi_initialized = 0;` and also call `cg_set_file_type(CG_FILE_HDF5));`.  This needs to be done in a more user-friendly manner, but works OK in the proof of concept.

Comments?  Is this worth pursuing, or did I overlook something that is a deal-breaker for supporting this behavior?


MicK7:
Is this POC complementary to the recent merge of #105 which allows usage of serial read/write in parallel context ?

gsjaardema:
@MicK7 
I haven't looked at #105 in detail, but based on a cusory glance it may be the #105 could be the path to removing the kluge that I mention in the second to the last paragraph of my original comment.  I will take a closer look and see how my patch interacts with #105

gsjaardema:
OK, I think I have convinced myself that this is not necessary.  Setting the communicator to `MPI_COMM_SELF` prior to calling `cgp_open()`  seems to give the correct behavior.

PR #62, opened 2018-04-25
Improve error messages, fix error check

gsjaardema:
Improved the error messages.  The original message for checking for valid range would output something that didn't make sense or at least wasn't helpful.  Split this into two checks and output info that helps locate problem.

Added `connectname` to error messages to show which connection had issues.

The transform check was testing `abs(transform[i]) < 0` which isn't possible.  It should be `<=0` or `<1`


brtnfld:
CGNS-136: Improve error messages, fix error check #62

PR #65, opened 2018-06-06
Fix Windows build with cmake

NovA80:
The patch addresses the following issues on building CGNS in Windows 7 x64 using CMake-3.11.3 & Visual Studio 2017.
1) Zlib library path specification in cmake was failing.
2) open64() syscall is not available on Windows.


brtnfld:
I made minor modifications to the PR and checked it in. Thanks

PR #67, opened 2018-06-20
support new ngon layout

MicK7:
- keep existing functions for FIXED_SIZE elements
- add functions cg_poly_* for poly elements section (NGON, NFACE, MIXED)
- add a converter tool "update_ngon" to convert old files
- change MLL version in order to dynamically load and update older files.
- update tests to support new layout.


MicK7:
It seems the continuous integration robot found a problem with the patch on Windows for HDF-1.8.
The crash must be very specific. Can anyone reproduce the bug ?

MicK7:
Code Review is welcomed now.

PR #69, opened 2018-08-06
Build parallel CGNS on Windows

mennodeij:
I ran into a problem where CGNS with parallel support would not build on Windows (using MS-MPI and parallel HDF5), because 'variable length arrays' are not supported in Visual C. Therefore, these have been replaced with malloc/free version to fix this problem.

### DIFF
--- a/src/pcgnslib.c
+++ b/src/pcgnslib.c
@@ -69,9 +69,14 @@ static int readwrite_data_parallel(hid_t group_id, CGNS_ENUMT(DataType_t) type,
 {
   int k;
   hid_t data_id, mem_shape_id, data_shape_id;
-  hsize_t start[ndims], dims[ndims];
+  hsize_t *start, *dims;
   herr_t herr, herr1;
   hid_t type_id, plist_id;
+
+  start = (hsize_t*)malloc(ndims * sizeof(hsize_t));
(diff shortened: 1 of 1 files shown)


MicK7:
Wouldn't it be easier to just have a fixed size like this :
hsize_t start[**CGIO_MAX_DIMENSIONS**], dims[**CGIO_MAX_DIMENSIONS**];

mennodeij:
Maybe, but I am not sufficiently on top of the code to see if that would make a difference. My proposed implementation is functionally identical. CGIO_MAX_DIMENSIONS is defined to be 12. 
If you proposal will "always work" I'm happy to change it, as it is indeed much easier.

MicK7:
The simple version is supposed to be thread safe and faster. Since it only consumed 12 x 2 x sizeof(hsize_t), I assume it is not very big to fit in the stack. I can't think of situation where it would fail. I agree that your implementation is functionally identical.
@brtnfld @gsjaardema Is there a pratical test for this function that could help decide the good solution ?

mennodeij:
Apologies for the delay in getting this done, I have been out of the office for a while. The method of using CGIO_MAX_DIMENSIONS has been adopted, making the change rather trivial.

PR #70, opened 2018-08-10
Fix bad if test --

gsjaardema:
An if test is using `=` instead of `==`


MicK7:
Thanks for catching this one. 
Probably by dint of reading the code, I do not see the missing "=" anymore.

PR #78, opened 2018-11-01
remove some gcc warnings

MicK7:
- remove a printf used for debug
- specify max string length to prevent buffer overflow in sprintf

### DIFF
--- a/src/cgns_internals.c
+++ b/src/cgns_internals.c
@@ -7039,7 +7039,7 @@ int cgi_write_model(double parent_id, cgns_model *model)
     }
 
      /* xModel_t */
-    sprintf(label,"%s_t",model->name);
+    sprintf(label,"%.30s_t",model->name);
     dim_vals = (cgsize_t)strlen(ModelTypeName[model->type]);
 
     if (cgi_new_node(parent_id, model->name, label, &model->id,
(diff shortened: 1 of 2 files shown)


src/cgns_internals.c
```diff
@@ -7039,7 +7039,7 @@ int cgi_write_model(double parent_id, cgns_model *model)
     }
 
      /* xModel_t */
-    sprintf(label,"%s_t",model->name);
+    sprintf(label,"%.30s_t",model->name);
```
smguzik: Suggest snprintf(label, 33, model->name);

brtnfld: I don't have a preference for snprintf over sprintf, so it's up to you. CGNS only uses snprintf once and it is in the tests.

smguzik:
It better describes intent.  %30 looks like a formatting decision whereas snprintf looks like a memory control (which it is).  Either way, can you merge that pull request?  None of us have permission.  This is indeed an overflow that should be corrected.

On 11/05/2018 08:00 AM, Scot Breitenfeld wrote:

@brtnfld commented on this pull request.

________________________________

In src/cgns_internals.c<https://github.com/CGNS/CGNS/pull/78#discussion_r230805663>:

> @@ -7039,7 +7039,7 @@ int cgi_write_model(double parent_id, cgns_model *model)
     }

      /* xModel_t */
-    sprintf(label,"%s_t",model->name);
+    sprintf(label,"%.30s_t",model->name);


I don't have a preference for snprintf over sprintf, so it's up to you. CGNS only uses snprintf once and it is in the tests.

PR #79, opened 2018-11-02
Rind plane indexing

MicK7:
Add cg_array_general_read/write


MicK7:
@smguzik is this implementation of cg_array_general_xxx what you had in mind ?

PR #80, opened 2018-11-03
Fix broken test with Rind Plane Indexing

MicK7:
Following the removal of grid number as a  parameter for cg_coord_general_xxx, tests were failing.

As a side note, deforming meshes using multiple GridCoordinates nodes won't be readable or writable with rind planes with the new API since cg_array_general_read/write is not aware of Rind Indexing thus always falling back to a CG_CONFIG_RIND_ZERO state (as it was done before).

### DIFF
--- a/src/cg_ftoc.c
+++ b/src/cg_ftoc.c
@@ -243,14 +243,14 @@ CGNSDLL void FMNAME(cg_add_path_f, CG_ADD_PATH_F) (STR_PSTR(pathname),
 
 CGNSDLL void cg_set_rind_zero_f(cgint_f *ier)
 {
-    *ier = (cgint_f)cg_configure(CG_CONFIG_RIND_INDEX, CG_CONFIG_RIND_ZERO);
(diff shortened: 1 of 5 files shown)


smguzik:
The reason that I removed that was because it was not consistent with other cg_coord_write.  The interfaces: cg_coord_write, cg_coord_partial_write, cg_coord_general_write provide increasing granularity in how you can access data: 1-to-1, hyperslab-in-file, hyperslab in memory and file.

I don't know what to do about deforming meshes but I'd prefer a more consistent fix across all three levels.  For now, I don't think we are breaking anything as long as the 1-to-1 writes can be performed with rind planes and deforming meshes.

smguzik:
I'm also reverting the change in cgns_internals.c since it's not related to Rind_Plane_Indexing and I am unsure of the impact.

smguzik:
Ah, now I see.  snprintf is probably a better fix and should go into develop.

PR #82, opened 2018-11-12
Remove double condition test

MicK7:
allow_dup condition is already checked on the previous line


smguzik:
Those are separate if statements so the condition needs to be checked on both lines.

PR #85, opened 2018-11-15
fix tests failing when compiled for 64bit

MicK7:


smguzik:
Thanks for the suggestions.  Fixes have been added in commit 093a614 with a slightly different strategy for Fortran (32-bit arguments in contained functions and returning 64-bit).

PR #86, opened 2018-12-03
Fix parallel write

MicK7:
In parallel writing mode NULL pointer can be used to write node without data


brtnfld:
Is this PR separate from what Stephen is working on, or is that work going to be a different PR?

smguzik:
Merged

PR #87, opened 2018-12-04
CGNS-146 (#87)

zbeekman:
* CGNS-146 :: Can't compile using MSVC & IVF on windows (MSVS 14)

Explicitly set the CMake `LINKER_LANGUAGE` property on `cngs_static` to `C`.
This is a work around for a potential CMake bug, and was suggested by
@bradking. Here is the issue and merge that introduced the bug:

 - https://gitlab.kitware.com/cmake/cmake/issues/16738
 - https://gitlab.kitware.com/cmake/cmake/merge_requests/643

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -608,7 +608,7 @@ set(headers
 
 if (CGNS_ENABLE_FORTRAN)
   list(APPEND headers
-    ${CMAKE_CURRENT_BINARY_DIR}/cgns.mod)
+    $<TARGET_FILE_DIR:cgns_static>/cgns.mod)
 endif (CGNS_ENABLE_FORTRAN)
 
 if (CGNS_ENABLE_PARALLEL)
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -612,8 +612,13 @@ set(headers
   ${CMAKE_CURRENT_BINARY_DIR}/cgnsBuild.defs)
 
 if (CGNS_ENABLE_FORTRAN)
-  list(APPEND headers
-    $<TARGET_FILE_DIR:cgns_static>/cgns.mod)
+  if(DEFINED CMAKE_Fortran_MODULE_DIRECTORY)
+    list(APPEND headers
+      ${CMAKE_Fortran_MODULE_DIRECTORY}/cgns.mod)
+  else()
+    list(APPEND headers
+      $<TARGET_FILE_DIR:cgns_static>/cgns.mod)
+  endif()
 endif (CGNS_ENABLE_FORTRAN)
 
 if (CGNS_ENABLE_PARALLEL)
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -532,6 +532,8 @@ if (CGNS_ENABLE_FORTRAN)
 else (CGNS_ENABLE_FORTRAN)
   add_library(cgns_static STATIC ${cgns_FILES})
 endif (CGNS_ENABLE_FORTRAN)
+# Needed to work around a CMake > 3.8 bug on Windows with MSVS and Intel Fortran
+set_property(TARGET cgns_static PROPERTY LINKER_LANGUAGE C)
 
 
 # Build a shared version of the library
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -236,6 +236,11 @@ CHECK_FORTRAN_FEATURE(Fortran2008
            END FUNCTION bind2
          END INTERFACE
        END MODULE test
+
+       PROGRAM main
+         USE test
+         IMPLICIT NONE
+       END PROGRAM
   "
   CHECK_FORTRAN_2008
 )


zbeekman:
@brtnfld I think this should completely fix CGNS-148 :: CGNS fails to compile on Windows with Fortran enabled

There appears to be a CMake bug/regression that was introduced after 3.8. I'm working on testing locally now, I'll report back once I have confirmed a successful build under MSVS on Windows with Fortran enabled.

zbeekman:
FYI, I had the wrong issue number & title so I amended the commit and force pushed it. I hope that doesn't cause problems for CI.

zbeekman:
I can confirm that the build now succeeds on Windows under MSVS with Fortran enabled.

A corresponding issue has been opened on teh CMake bug tracker: https://gitlab.kitware.com/cmake/cmake/issues/18687

A minor problem remains: The installation of the cgns `.mod` file fails due to the way MSVS behaves. The CMake installation is not looking under the `$<CONFIG>` subdirectory for the module file, and thus the installation of the .mod file is failing. I'll try to add a fix for that here as well.

brtnfld:
Should I wait to merge this until the .mod file issue is resolved?

brtnfld:
Shouldn't you add conditionals for Windows around these new statements? It works correctly on LInux and OSX.

zbeekman:
I’m testing the mod file fix on Windows right now. The new property on the
object library can and should be applied unconditionally, no reason to make
it a special case for windows unless it’s somehow breaking macOS/Linux. The
property on the object library is a CMake bug work around. The value is
correct regardless of OS.

The install mod file fix should also be generator/OS agnostic but I am
still testing that.

src/CMakeLists.txt
```diff
@@ -532,6 +532,8 @@ if (CGNS_ENABLE_FORTRAN)
 else (CGNS_ENABLE_FORTRAN)
   add_library(cgns_static STATIC ${cgns_FILES})
 endif (CGNS_ENABLE_FORTRAN)
+# Needed to work around a CMake > 3.8 bug on Windows with MSVS and Intel Fortran
```
zbeekman: This comment may be misleading: I'm not 100% sure if the bug is unique to Windows/Intel Fortran. A regression was introduced in CMake that triggers this. Effectively, enabling Fortran caused the linker language for `cgns_static` to change (to Fortran I think) due to the `OBJECT` library `cgns_f2c`. This happens in newer CMakes.

@brtnfld if you want me to adjust the comment please let me know.

src/CMakeLists.txt
```diff
@@ -606,7 +608,7 @@ set(headers
 
 if (CGNS_ENABLE_FORTRAN)
   list(APPEND headers
-    ${CMAKE_CURRENT_BINARY_DIR}/cgns.mod)
+    $<TARGET_FILE_DIR:cgns_static>/cgns.mod)
```
zbeekman: This change should be safe on all platforms, with all configurations. For multi-config generators, like MSVS, targets get put into a subdirectory with the `$<CONFIG>` name. By piggybacking off of the location of the `cgns_static` directory we can determine where the compiler put the .mod file, assuming, users aren't doing anything crazy like setting [`CMAKE_Fortran_MODULE_DIRECTORY`](https://cmake.org/cmake/help/v3.13/variable/CMAKE_Fortran_MODULE_DIRECTORY.html).

If you like, I could add a check for this, and if set, then use that location instead. This would be fairly pathological on the part of the user, but you never know.

zbeekman:
@brtnfld FYI, I noticed that the introspection for `HAVE_FORTRAN_2008` is probably broken on most/all systems. It's possible that older CMakes passed `-c` during `try_compile()` but for me with GFortran and ifort, the compiler tries to link the code, and since the test code only defines a module, the linking done by the compiler fails with error messages about missing `formain` etc. Just adding a mostly empty `program main` ... to the bottom of the test code should resolve this.

I have an additional commit that I could push, if you like, that would take care of this.

src/CMakeLists.txt
brtnfld: We should probably allow for user's changes.

brtnfld:
I think we have had issues with newer cmake, mainly on windows. Though I can't find a bug report for it. I would just add a main program.

brtnfld:
This also looks related to 
https://cgnsorg.atlassian.net/browse/CGNS-143?filter=-5
Correct?

zbeekman:
> This also looks related to
> [cgnsorg.atlassian.net/browse/CGNS-143?filter=-5](https://cgnsorg.atlassian.net/browse/CGNS-143?filter=-5)
> Correct?

The logic behind the library naming scheme and choosing shared vs static libraries eludes me somewhat, and is very non-canonical CMake, IMO. Typically, projects either:

a) Rely on `-DBUILD_SHARED_LIBS:BOOL=<ON|OFF>` to target shared or static libraries, or
b) Explicitly pass `STATIC` or `SHARED` to all invocations of `add_library()`

Further confusing me is the fact that I think the comment in https://cgnsorg.atlassian.net/browse/CGNS-143?filter=-5 has an instance of `SHARED` and `STATIC` swapped:

> However, I also had to fix the src/CMakeLists.txt, because it assumes the default library type is SHARED, and it is not when BUILD_SHARED_LIBS is ON.

Should read

>However, I also had to fix the src/CMakeLists.txt, because it assumes the default library type is STATIC, and it is not when BUILD_SHARED_LIBS is ON.

if I am following the conversation correctly.

The commenter's proposed solution of setting `-DBUILD_SHARED_LIBS:BOOL=ON` seems out of line with the way the CGNS build system is currently structured.

TL;DR: This might be related, but it's certainly not directly related, AFAICT. Opening the log file attached to the issue shows mostly undefined references during linking. This makes me think that the additional comment may be a bit of a red herring. Maybe the API changed in HDF5 or maybe something caused the linking order to be different. At any rate, CGNS-143 is far beyond the scope of this pull request, in my opinion.

PR #90, opened 2019-01-29
Rind plane indexing

MicK7:
Rind plane indexing implémentation according to CPEX #40


gsjaardema:
Just noticed that following this patch, CGNS now requires C99 mode for compilation since there are several instances of `for (int n=0; n < ...; n++)` added in this patch.  Not sure if that is a bad thing or not, but wanted to make sure it was an intentional change.

I think there may need to be a change to the CMake to select C99 mode...

smguzik:
I think the C99 features can help create cleaner, more understandable code and with less errors.  I don't know what the reasons would be for keeping C89 compatibility.  I think we should conservatively advance the standard and C99 is 2 decades old.

gsjaardema:
Note that in the cases in this file, `n` is already defined, so the `int n` in the for loop is shadowing the outer n.  In this case, it is probably best to clean it up and remove the `int n` in the for loops.  I have another PR which does this #95

gsjaardema:
I also agree that being able to use C99 or later would be nice and useful and more robust normally.  However, we then need to add some CMake code to select C99 compiler options and I think that requires that we require a newer version of CMake or the adding of additional CMake code.

smguzik:
I think we do need to add the compiler option.  A lot of the code I wrote in the rind_plane_indexing branch mixes declarations with code and uses single assignment.  Formally it is c99 but I think most c89 compilers support it as an extension.  We should either rewrite the code to strict c89 or force c99.

Rewriting the rind_plane_indexing branch to c89 would be an exercise in making the code less readable and robust.  For example, "const int access_full_range =..." gives assurance that the variable is initialized and will (or at least should) not change in the encompassing scope.  So I'm in favor of c99.  Also,  if I force c89, the code does not build for other reasons.  But c99 passes.

We should run this through the committee though.

PR #92, opened 2019-01-30
Move assignment to pointer after NULL check

gsjaardema:
The `found` pointer is assigned to prior to checking whether it was NULL.  Move assignment after check.

This change is large since github decided to normalize line endings...

### DIFF
--- a/src/adf/ADF_internals.c
+++ b/src/adf/ADF_internals.c
@@ -4885,9 +4885,6 @@ double     root_ID = 0;
 int        i ;
 
 
-*error_return = NO_ERROR ;
-*found = 0;
-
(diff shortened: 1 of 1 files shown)


brtnfld:
Your file adds the carriage-return character, probably from Windows. Can you remove those characters?

gsjaardema:
I can’t change it.   It was added automatically by github when I edited the
file in github.   I think it had Windows endings and changed to Linux maybe?

.. greg

smguzik:
It looks like it did have Windows endings originally.  Maybe we should run a fromdos first and commit that before a code change.

FYI, one can use
find . -name "*.c" -print0 | xargs -0 grep -l '^M$'
to find these files where ^M is not two separate characters.  Use ctrl+v+m to enter this character.

PR #95, opened 2019-02-05
Eliminate C99 style for loop -- unneeded

gsjaardema:
Eliminate for loops of the form `for (int n=....` . In this case, the only change needed is to remove the `int` since `n` is already declared in the routines and the `int n` is just causing a shadowing of the previously defined `n`.  Also eliminates need to use C99 or later standard when compiling.


brtnfld:
This fixes issue on eclipse (SNL), and intel/16, openmpi-1.10

PR #100, opened 2019-02-12
 Fixed compliation error with cc: Sun C 5.12 SunOS_sparc

brtnfld:


smguzik:
Can we use ISO C99 with Sun compilers?

brtnfld:
Maybe it is a compiler bug, but for the sun compiler if the  _XOPEN_SOURCE 500 is placed after the stdio.h include it works. But if it is before it, then I get the error 

"cgns_io.c", line 216: incomplete struct/union/enum stat64: st
"cgns_io.c", line 224: improper member use: st_mode
"cgns_io.c", line 226: improper member use: st_size
"cgns_io.c", line 229: improper member use: st_size
"cgns_io.c", line 230: improper member use: st_size
"cgns_io.c", line 577: incomplete struct/union/enum stat64: st
"cgns_io.c", line 581: warning: implicit function declaration: stat64
"cgns_io.c", line 582: improper member use: st_mode

Line 216 is 
#ifdef S_IFLNK
    struct stat st;
#endif

I've reverted the change for now. The sun compiler is full C99.

PR #106, opened 2019-02-18
Eliminate potential integer overflow / undefined behavior

gsjaardema:
The check for open files was summing the `g_files` entry when all that was needed was a yes/no on whether there were any open files.  Depending of the value of `g_files` it was possible for the integer sum to overflow the range of a signed int which gives undefined behavior.  `idx` is an integer and `g_files` is an `hid_t` which is 64-bits on recent HDF5 versions, so overflow is definitely a possibility (and has happened in the wild)


gsjaardema:
Note that this change has existed in #61 for awhile but splitting it into its own PR since it bit me again when I was running with a "vanilla" CGNS.  Hopefully if in its own PR, it will get merged.

PR #114, opened 2019-02-21
Create hdf5-1.8.x compatible files from 1.10.X library

gsjaardema:
If CGNS is built against the new HDF5-1.10.x library versions, the files that are created are not readable by applications which are built against earlier HDF5-1.8.x library versions.  This can cause issues on many systems which are using commercial or externally maintained visualization or other applications (ParaView, Ensight, ...) . 

Adding the command to use the `H5F_LIBVER_V18` option makes the files readable by the older applications.  

Here is some advice from the HDF5 group related to this issue:
```
If an application written for 1.8.* is linked with HDF5 1.10.(i.e., it doesn’t use any new features of 1.10.0), a file created by the application will be readable by HDF5 1.8. (or even earlier versions). The exception to this rule would be an application that uses H5Pset_libver_bounds with H5F_LIBVER_LATEST value for its “low” and “high” parameters.

In this case both parameters should be set to H5F_LIBVER_V18 to create 1.8 compatible file.

File control versioning became available in the HDF5 1.10.3 release. As a safeguard feature, we recommend using H5Pset_libver_bounds function to control file versioning.
```

### DIFF
--- a/src/adfh/ADFH.c
+++ b/src/adfh/ADFH.c
@@ -2061,7 +2061,11 @@ void ADFH_Database_Open(const char   *name,
   /*  H5Pset_latest_format(fapl, 1); */
   /* Performance patch applied by KSH on 2009.05.18 */
   H5Pset_libver_bounds(g_propfileopen,
+#if H5_VERSION_GE(1,10,3)
+		       H5F_LIBVER_V18, H5F_LIBVER_V18);
+#else
 		       H5F_LIBVER_LATEST, H5F_LIBVER_LATEST);
+#endif
   /* open the file */
 
 #ifdef BUILD_PARALLEL


MicK7:
Can it be made optional at compile time with a flag KEEP_COMPATIBILITY_HDF5_V18 that would be ON but can be turned off ?

gsjaardema:
@MicK7 What benefit would that provide.  In other words, is there a benefit that using the V110 file format currently provides?

I have no problem with it being optional other than the added complexity it would cause and the need for additional build-related documentation.

MicK7:
It would be easier for people who want to work with HDF5 1.10 to experiment.
I was also thinking that it could be useful for metadata compression work or collective metadata IO, not sure though.

gsjaardema:
Fair enough.  I think that the collective metadata still works on the hdf5-1.8.X format is using the hdf5-1.10.x library.  I'm not sure whether the upcoming `PROC0` and other scalability-related changes would be usable.  Do you know @brtnfld  ?

brtnfld:
This is scheduled to be fixed in 3.4.1:
https://cgnsorg.atlassian.net/browse/CGNS-138

I would suggest adding CG_FILE_HDF5_1_8 CG_FILE_HDF5_1_10 constants that can be used by
cg_set_file_type

1.8 written files should still be able to use the proc0 and other scalability changes. These changes are not associated with the HDF5 file format.

gsjaardema:
Is 3.4.1 the upcoming release?

brtnfld:
The release next release is 3.4.0, if you want me to push this feature to 3.4.0 let me know.

gsjaardema:
If possible, it would help a lot.  We have people that won't upgrade from hdf5-1.8.x since they need to use Paraview and other tools which haven't moved to 1.10.x yet so their files become unreadable if they use hdf5-1.10.x.  However, they are running into scalability issues at large processor counts with hdf5-1.8.x

If it doesn't make it into 3.4.0 I can provide a local patch file, so not disastrous, but inconvenient.
Thanks.

MicK7:
Concerning ParaView, I am in the process of migrating the reader to CGNS 3.4 and it is already possible to compile with 1.10 support (if it is not the case it should or thus I will need to create a ticket issue).

brtnfld:
I'm thinking we should set the default to using HDF5 1.8 format. CGNS does not use any 1.10 features that are associated with the 1.10 format (i.e. VDS, chunking, etc...). When these features are introduced into CGNS, then we can default to use the 1.10 format.

I don't think this will hurt performance, but I have not checked to be sure.

gsjaardema:
I have done some non-rigourous timings and I didn't see any major changes in performance with 1.8.X compatible files vs 1.10.X files.  Again, this was not well-controlled or rigourous.  Just a few runs with both file formats on a wide range of processors.

brtnfld:
Created PR #119.

brtnfld:
Compact storage introduces the use of:
H5Pset_file_space_strategy
which is a 1.10 format. So we will have to introduce a method of specifying which HDF5 library version to use.

brtnfld:
The HDF5 library (develop) was updated to throw an error if H5Pset_file_space_strategy(…) is specified with 1.8 version compatibility. I’ve disabled the option for now until the issue is resolved on how to determine what version of HDF5 should be used. We will need to introduce an option or API which lets users specify this, CGNS-166. I'll create an RFC summarizing the possibilities.

PR #115, opened 2019-02-22
CGNS-129 Fix to build shared lib on windows.

matmel:
This pull request fixes the cmake build system on windows. Hopefully, from now on we should be able to build the static and shared versions of the library and tests without error. If the shared version of CGNS is chosen, the shared version of HDF5 will be automatically used. I updated the appveyor build matrix to add a shared build.

There is a limitation though: we cannot build the shared and static versions at the same time. We could improve cmake scripts to support it if desired. This is a more involved work though.


brtnfld:
Tony, can you make sure this is compatible with Intel (serial and parallel) on windows? I need to renew my Intel Windows 19 License or else I would do it.

brtnfld:
Tentatively approving, we need to verify it works for intel on windows before we release it (this is on-going).

brtnfld:
We will have to test your fork offline before merging it again. I don't think I will have time to investigate it before the release of 3.4.0.

PR #124, opened 2019-03-06
Unify similar defines

gsjaardema:
The code contains a couple similar defines: `BUILD_HDF5` and `CG_BUILD_HDF5` and also the `BUILD_PARALLEL` and `CG_BUILD_PARALLEL`.  The "non CG_" versions are set as compile time defines and their values cannot be queried after the library has been installed to determine their values -- the value is only available at build time.

The `CG_BUILD_*` versions do mirror their values, so they can be queried; however, there can be other values inside an include file that are inside if blocks which use the "non CG_" versions.  For example, it is not possible to determine whether an installed CGNS library had the `HDF5_HAVE_COLL_METADATA` define enabled at build time.  Since the absence/presence of this define can have an effect on parallel scalability, it would be nice to be able to determine whether it was set at build time.

While looking at making this setting visible, I stumbled upon the `CG_BUILD_*` and `BUILD_* `similarities and it took awhile to figure out where the various setting were and why they were different.  In an effort to perhaps simplify things and also make the `HDF5_HAVE_COLL_METADATA` visible after installation, I created this PR.  

I think it captures everything and gives the same build as prior to the change, but I don't build Fortran, so the change in the fortran file is just a guess.

I'm not real invested in this PR, so if it causes more issues than it solves, i am fine with rejecting it.  My main issue it that it would be nice to be able to determine the setting of the collective metadata define in a library I am linking to.


MicK7:
I stumbled upon the same issue with CG_BUILD_HDF5 while linking inside ParaView.

brtnfld:
I don't see anything wrong with the PR, looks like a good change.

PR #125, opened 2019-03-07
Fix bad assert

gsjaardema:
Assert statement was assigning instead of checking for equality.


brtnfld:
Are you addressing Coverity issues? If so, then you can work with the coverity_scan branch (you can push to it). We can then merge those fixes to develop in a single PR. I synced the branch with develop this morning.

gsjaardema:
OK. Sounds good.  I had some time during a compile and it looked like an easy fix, but will do those on the branch in the future.

PR #130, opened 2019-05-13
Compact storage rev

brtnfld:
Added HDF5 compact storage to CGNS to improve parallel IO performance. The default storage is compact storage unless the dataset does not meet the < 64KiB limit. Also, if the dataset can have partial IO, then it is contiguous storage. Also, removed 'tab' spacing. Everything should be transparent to the application code.


MicK7:
Looks good. Although some indentation changes do not seem necessary.
Is a user documentation about configuration flags planned ?

brtnfld:
I did not introduce a new configuration option. What are you referring too?

MicK7:
Sorry. I thought that HDF5_HAVE_FILE_SPACE_STRATEGY would have been configurable by the user. But It seems it is automatically set depending on HDF5 configuration and it is way better like this.
Thanks for the good job.

PR #137, opened 2019-06-17
Do not read elements when computing partial data size

MicK7:
To reduce memory usage of cg_ElementPartialSize only load ElementStartOffset array and compute ElementDataSize from it


MicK7:
This is a simple fix to mitigate an excessive memory usage. May be it can be improved by doing also a partial load of ElementStartOffset array

PR #138, opened 2019-07-18
Fix HDF5 feature detection

mennodeij:
I have a build of HDF5 1.10.5 with `HDF5_ENABLE_PARALLEL=ON`

When I then request CGNS to use HDF5 with parallel support using the following cmake configuration
```
-DCGNS_ENABLE_PARALLEL=ON
-DCGNS_ENABLE_HDF5=ON
-DHDF5_NEED_MPI=ON
-DHDF5_DIR=<path-to-hdf5-install>/share/cmake/hdf5
```
, I see that the feature detection of the HDF5 features `Collective_metadata` and `H5Pset_file_space_strategy` are both failing, whereas these are available when using `./configure` against the same build of HDF5.

To correct this the attached changes to the `CMakeLists.txt` file were made. With these changes, the features are correctly detected.


mennodeij:
Thanks!

PR #139, opened 2019-08-15
Make static mpi-related variables consistent

gsjaardema:
Currently, it is possible for the `pcg_mpi_comm` to be inconsistent with `pcg_mpi_comm_rank` and `pcg_mpi_comm_size`.  
The latter two values are set during a call to `cgp_open` and the first is set during a call to `cgp_mpi_comm`.  If the user is using both parallel and serial file access during the same run, they may call `cgp_mpi_comm` multiple times and the values of `pcg_mpi_comm_rank` and `pcg_mpi_comm_size` will only be consistent with the communicator in `pcg_mpi_comm` if the user calls `cgp_open`.  

With the changes here, all three values (and the `ParallelMPICommunicator`) will always be consistent.  Also, if the user does not call `cgp_mpi_comm()` before calling `cgp_open()`, this will be detected and everything will be consistent.

### DIFF
--- a/src/pcgnslib.c
+++ b/src/pcgnslib.c
@@ -366,11 +366,22 @@ static int check_parallel(cgns_file *cgfile)
 
 int cgp_mpi_comm(MPI_Comm comm)
 {
-    pcg_mpi_comm=comm;
-    if( cgio_configure(CG_CONFIG_HDF5_MPI_COMM, &comm) != -1) {
-      return CG_ERROR;
+    /* check if we are actually running a parallel program */
+    /* Flag is true if MPI_Init or MPI_Init_thread has been called and false otherwise. */
+    pcg_mpi_initialized = 0;
+    MPI_Initialized(&pcg_mpi_initialized);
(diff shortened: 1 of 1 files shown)


gsjaardema:
This PR also checks that mpi has been initialized in `cgp_mpi_comm()` and returns the correct error code if it hasn't been initialized.

brtnfld:
Is this to address the issue of creating a file in serial, opening in parallel, closing, and opening in serial?

brtnfld:
I added more tests to ptest/open_close.c, your changes fixes the case of opening the file with COMM_SELF, which fails in 3.4.0. See test.

gsjaardema:
@brtnfld I am able to open a file in parallel for read and serial for write  simultaneously (and vice-versa) with or without this change using the `MPI_COMM_SELF` suggestions...  Not sure why it isn't working for your use case.

brtnfld:
Works after I syncing my fork.

PR #140, opened 2019-08-15
Use correct MPI communicator

gsjaardema:
The `MPI_Allreduce` call in `cgp_parent_data_write()` is using `MPI_COMM_WORLD`  It would be more correct to use the communicator stored in `cgp_mpi_comm` which is set by the client. If the client is not using `MPI_COMM_WORLD` and calls this function, it will hang since not all ranks will be participating in the call.  

There is another use of `MPI_COMM_WORLD` in `cgp_error_exit()`, but I'm not sure if there is a use case for keeping that as is (I seem to remember some discussion a year or so ago)


brtnfld:
The parallel test fails:
test_unstruc_quad ...  *** FAILED *** 

I checked that it passes in develop

PR #146, opened 2019-10-16
remove warning from -Wunused-result

MicK7:
on travis two warnings keep appearing, this commit remove them


MicK7:
This is a small general maintenance commit

PR #154, opened 2020-01-22
CPEX0045 High order

LaurentCadet:
The work related to the implementation of the **CPEX045** (high order CGNS) is done.

This implementation comes with the following tests/examples :
- src/tests/test_high_order.c
  - _general MLL tests_
- src/ptests/para_high_order.c
  - _example of implementation for parallel CGNS according to CPEX045_
- src/Test_UserGuideCode/C_code/write_grid_ho_unst.c
  - _write an high ordered Mesh (2nd order QUAD with family and lagrange control points)_
- src/Test_UserGuideCode/C_code/write_flowcell_ho_unst.c
  - _add a 3rd order solution to the mesh created by write_grid_ho_unst.c_
- src/Test_UserGuideCode/C_code/read_flowvert_ho_unst.c
  - _check that the solution order is correctly written in the file_

The branch has been updated wrt _develop_ and the _ctest_ is clear.

I am available for any questions or fixes.

Thanks for waiting,

Laurent CADET.


src/tools/cgnscheck.c
```diff
@@ -27,6 +27,8 @@
 
 #define USE_MID_NODES
 
+#define min(A,B) A <= B ? A : B;
```
MicK7: min and max should be already located in cgns_header.h

src/ptests/para_high_order.c
```diff
@@ -0,0 +1,334 @@
+/*
+! @file para_high_order.c
+! @author Laurent Cadet <laurent.cadet@cenaero.fr>
+! @version 0.1
+!
+! @section LICENSE
+! BSD style license
```
MicK7: CGNS software should use CGNS license (zlib/libpng modified license)

src/cgns_internals.c
```diff
@@ -40,6 +40,9 @@ freely, subject to the following restrictions:
 
 #define CGNS_NAN(x)  (!((x) < HUGE_VAL && (x) > -HUGE_VAL))
 
+#define max(a,b) (a>=b?a:b)
```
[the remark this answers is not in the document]
LaurentCadet: Mistakes. I will correct in the next commit

src/tools/cgnscheck.c
LaurentCadet: Mistakes. I will correct in the next commit

src/ptests/para_high_order.c
LaurentCadet: This testcase is a copy of the src/ptests/test_unstructured.c. Thus its header is the same ...
should I correct all of them ?

src/cgns_header.h
```diff
@@ -233,6 +233,8 @@ typedef struct {            /* IndexArray/Range_t Node      */
     CGNS_ENUMT(PointSetType_t) type;  /* PointList, PointRange, ...       */
     char_33 data_type;      /* type of data                         */
     cgsize_t npts;          /* number of points to define the patch */
+    cgsize_t range_min[12]; /* start of range if type is PointRange or ElementRange */
```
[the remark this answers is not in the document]
LaurentCadet: The aim was to avoid to read the points every times I would need the ranges. I will correct It in the next commit with the introduction of the function :
`int cgi_ptset_range(cgns_ptset *ptset, cgsize_t *range_min, cgsize_t *range_max)`

src/cgns_internals.c
```diff
@@ -49,6 +52,7 @@ extern int HDF5storage_type;
 int Idim;           /* current IndexDimension          */
 int Cdim;           /* current CellDimension           */
 int Pdim;           /* current PhysicalDimension           */
+cgns_zone *CurrentZonePtr; /* current Zone structure pointer */
```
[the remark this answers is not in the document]
LaurentCadet: Due to High order specificities, It is required to acces the Element sections in the solution data size method. I did it to avoid modifying too deeply the existing methods. However, I could modify the `cgi_read_sol()` to pass the zone as argument ...

brtnfld:
It looks like we lose forward compatibility with this feature and the major version number will need to be changed. Can you please verify this is true?

For reference: 

Q: What do the CGNS Version numbers mean, and does the CGNS library maintain backward/forward compatibility?

A: The CGNS versions are currently numbered as follows: "Version x.y, Revision z", or "Version x.y-z". (However, the revision number is often left off, so you will typically only see "Version x.y".) The first number represents the "major" version number. Within this number, the library maintains forward compatibility. For example, Version 2.3 of the library can read a Version 2.5 CGNS file, but Version 1.y cannot necessarily read any Version 2.y (or later) file. A new "major" version number is assigned either when forward compatibility is lost, or else when there is a significant change made to the API. The second number is the "point release" number. It increments when there are relatively minor changes to the API, or with the addition of new features. The third number (the revision number) changes with bug fixes. Major releases and point releases are announced (via the Latest News page and via the CGNSTalk Discussion Group), whereas revisions are generally not announced. Note that CGNS always maintains backward compatibility: the most recent version of the library will be able to read all older versions CGNS files.

brtnfld:
Can you make this feature a branch in the repo for now? We need to do further testing outside of github, have more people review it, and we would like additional users to try it out before officially releasing it. It's a major update so we really want to test the feature before releasing it. Thanks for all your hard work.

PR #158, opened 2020-02-19
bumped version to 4.1 for next release

brtnfld:


brtnfld:
This might need to be version 5.0 due to cpex0045

MicK7:
I would rather not include cpex0045 in the next release.
Indeed, it needs more testing to check for bugs and forward compatibility issues.

brtnfld:
I was similarily thinking, as I was hoping to release 4.1 with all the current changes in develop.
What I will suggest is the cpex0045 be added as a branch, for now, so that people can try it out before releasing it officially.

MicK7:
Ok, CPEX0045 can be talked over during next CGNS committee meeting.

PR #160, opened 2020-02-20
Update update_ngon.c

MicK7:


MicK7:
In the 4.0.0 it is partially wrong since it was modified to generate 4.0 files but the comparison against 4.0 was missed. Thus it does not lead to any unusable files.

PR #165, opened 2020-02-24
CMakeLists.txt: Disable RPATH setting if not using shared

sebrowne:
When not using shared libraries, do not do anything with RPATH on the
executables.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -345,23 +345,25 @@ endif (CGNS_ENABLE_HDF5 AND HDF5_NEED_MPI)
 # RPATH Management #
 ####################
 
-# use, i.e. don't skip the full RPATH for the build tree
-set(CMAKE_SKIP_BUILD_RPATH  FALSE)
+if (CGNS_USE_SHARED)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -345,23 +345,25 @@ endif (CGNS_ENABLE_HDF5 AND HDF5_NEED_MPI)
...
+  # which point to directories outside the build tree to the install RPATH
+  set(CMAKE_INSTALL_RPATH_USE_LINK_PATH TRUE)
 
-IF (${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
-   set(CMAKE_MACOSX_RPATH TRUE)
-ENDIF()
+  IF (${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
+     set(CMAKE_MACOSX_RPATH TRUE)
+  ENDIF()
+endif(CGNS_USE_SHARED)
```
MicK7:
```suggestion
else(CGNS_USE_SHARED)
set(CMAKE_SKIP_RPATH TRUE)
endif(CGNS_USE_SHARED)
```

PR #177, opened 2020-03-05
Test to fix CGNS96 PR

MicK7:


brtnfld:
I would like to merge PR #174 to remove the version number updates from the pull request.

brtnfld:
I wish we had tests for the cgnstools ...

There is also an open issue related to this:
https://cgnsorg.atlassian.net/browse/CGNS-147
But it looks like your changes still work with MPI. I will need a test on OSX. And we need to test cmake, and probably look into PR #14

MicK7:
@brtnfld I agree 😢 
The configure code is a big mess and I just did a quick and dirty fix.
The open issue https://cgnsorg.atlassian.net/browse/CGNS-147 is normal since the mpi detection code is commented in configure script...

MicK7:
I would prefer to improve PR#14

brtnfld:
Yes, which was why it was not merged, but we should address this issue when getting the cgnstools regression tested on other platforms.

PR #179, opened 2020-03-10
Handle integer size correctly

gsjaardema:
I think this fixes the problem of an integer mismatch between the size of `cgsize_t` and the size of the data stored in the `Elements` array in the CGNS file.  The data will be read from the file and stored in the `elements` array which is declared as `cgsize_t *elements`, so I think that should be the `type` that is passed do `readwrite_data_parallel`.  If `cgsize_t` is different than the size of the data in the file, then I think HDF5 will do the conversion.   I've tested this on the 4 possible settings (cgsize_t = 4 or 8, size of `Elements` on disk = I4 or I8) and they all work correctly with the fix.  Without the fix, I get a failure whenever the sizes don't match for a parallel read.

This *might* be needed for `cgp_elements_write_data` but I think the data type is set correctly in some other location currently.


gsjaardema:
Thanks.

brtnfld:
I also created an issue for testing reading older cgns files, and for different cgns builds, in the cgns parallel tests, which should test for this case:

https://cgnsorg.atlassian.net/browse/CGNS-200

PR #183, opened 2020-03-15
JIRA CGNS-194

MicK7:
Test update H5Giterate ...


MicK7:
It would be nice to also remove H5Gget_objinfo from ADFH.c

brtnfld:
> It would be nice to also remove H5Gget_objinfo from ADFH.c

Yes, I should have also included that because it is needed to run with DAOS. It should also be faster since we only need the basic metadata information.

MicK7:
@brtnfld Do I transfer the changes to a CGNS-194 branch on CGNS/CGNS so you can commit directly into it ?

brtnfld:
That will work.

PR #184, opened 2020-03-15
SU2 patch for Partial reading of offset array when computing element size

MicK7:
This is a backport of SU2 patch for CGNS (https://github.com/su2code/SU2)
It can surely be improved by using recent cgio_read_data_type API more efficiently.
I am not sure what the memory saving done by the patch is worth. Any feedback would be appreciated (this change makes the code ugly thus not easy to maintain)!


MicK7:
integrated in cgp_poly_experimental branch

PR #187, opened 2020-03-18
JIRA CGNS-194

MicK7:
Be fully HDF5 1.8 compliant

> Test update H5Giterate
> 
> We need to add a #if H5_VERSION_GE(1,12,0) for H5L_iterate_t; HDF5 1.12 introduces H5Literate_by_name2; a new type H5L_iterate2_t.
>
>It would be nice to also remove H5Gget_objinfo from ADFH.c
>
>Yes, I should have also included that because it is needed to run with DAOS. It should also be faster since we only need the basic metadata information.


MicK7:
H5Gget_objinfo can be replaced by H5Oget_info_by_name but it seems to be already obsolete since a H5Oget_info_by_name2 exists in recent hdf5 version.

MicK7:
It would be nice to check on a bigger case than the test if the changes do not introduce performance issue

brtnfld:
Agreed.

brtnfld:
Checking with our Daily Tests over the next couple of days.

MicK7:
@brtnfld any feedback from your daily testing ?

brtnfld:
Looks good.

PR #193, opened 2020-03-24
JIRA CGNS-38

MicK7:
https://cgnsorg.atlassian.net/browse/CGNS-38

### DIFF
--- a/src/configure
+++ b/src/configure
@@ -748,6 +748,7 @@ infodir
 docdir
 oldincludedir
 includedir
+runstatedir
 localstatedir
 sharedstatedir
 sysconfdir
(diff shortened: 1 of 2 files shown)


src/configure
```diff
@@ -4782,22 +4794,57 @@ $as_echo_n "checking cgsystem... " >&6; }
 $as_echo "$SYSTEM" >&6; }
   if test $do64bit = yes; then
     BUILD64BIT=1
-    is64bit=`echo $SYSTEM | grep 64`
-    test -z "$is64bit" && \
+    # The cast to long int works around a bug in the HP C Compiler
```
gsjaardema: Is it not possible to do this somewhat simpler using `AC_CHECK_SIZEOF(void *)` which sets`ac_cv_sizeof_void_p`.  It looks like autoconf's from about 2002 onward should already include the HP fix...

MicK7: Yes it is what is done in the configure.ac. But the configure changes come from running autoconf with an old autoconf I assume.

gsjaardema: OK, I see now.  Didn't click that this was the generated code.

PR #194, opened 2020-03-31
Upgrade minimum required version for cmake

MicK7:
Nowadays Hdf5 requires cmake 3.12. Thus a little upgrade to require at least a 3.8 version of cmake in CGNS library seems sane.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 2.8)
+cmake_minimum_required(VERSION 3.8)
 if(COMMAND cmake_policy)
   cmake_policy(SET CMP0003 NEW)
 endif(COMMAND cmake_policy)


brtnfld:
I don't think we should bump the requirement unless we actually require it due to a newer CMake feature (and there are some newer features that we should probably make use of). Some of the DOE machines don't have newer versions of CMake readily available and use autotools to build HDF5.

Although it is true that we don't test with older CMake versions, so we don't guarantee that older CMake versions will work.

MicK7:
CMake 3.8 is needed for compilation with Visual Studio according to Microsoft documentation. Indeed it introduced a server mode that is required by VS. Since this CMake is already 3 years old and that no older CMake is used for testing, it is a safe bet even for old machines that can find it in backports. Presently we do garantee that it works for DOE but since they use autotools, I don't think it will change anything.

PR #195, opened 2020-04-05
Small improvements

MicK7:
- Add environment variable to prevent cgnsview from locking CGNS/HDF5 files and let other tools like cgnscheck do their job
- Comment unlink in fortran test that creates problem on some platforms
- Rewrite initialization of stack array to remove GCC warning
- Add some missing CGNS library versions.
- Add a strcpy to prevent overflow with label shorter than ADF_NAME_LENGTH. (Internally it seems that label attribute has a fixed size of same length as the name attribute.)


MicK7:
Yes and maybe create a template Pull Request that lists what to check before release would be useful.
This is what guys from kitware do before releasing ParaView.

MicK7:
@brtnfld Thanks for reviewing

PR #196, opened 2020-04-18
Mapping X4 X8 for hdf5

MicK7:
Mapping of X4/X8 was lost when moving from ADF to HDF5.
This PR proposes to reintroduce it as an experimental feature.
It uses an H5T_COMPOUND and is compatible with standard C99 complex type.

The mapping is similar to what is done in other library like h5py.
Depending on application, people can still store each part of complex number in separate arrays by prefixing with Re or Im the name of the array.

Atlassian CGNS-202


MicK7:
Yes I renamed I in J before realising that I could just undef it. Thus their is no way that it could affect clients adversely since I also did the undef.

MicK7:
According to IBM xl documentation (https://www-01.ibm.com/servers/resourcelink/svc00100.nsf/pages/zOSV2R3sc147308/$file/cbclx01_v2r3.pdf), page 141, those lines are valid and should not fail. Which version of xl compiler does fail ?
Concerning PGI compiler, was the -c9x flag given to pgcc ?

brtnfld:
xl compiler, 15.1

-c9x was not used with the pgi compiler.

Sun compilers also have an problem:

"cgns_internals.c", line 5839: warning: implicit function declaration: __real__
"cgns_internals.c", line 5839: left operand must be modifiable lvalue: op "="
"cgns_internals.c", line 5840: warning: implicit function declaration: __imag__
"cgns_internals.c", line 5840: left operand must be modifiable lvalue: op "="
"cgns_internals.c", line 5853: left operand must be modifiable lvalue: op "="
"cgns_internals.c", line 5854: left operand must be modifiable lvalue: op "="

MicK7:
For xl compiler, the required options are:
-qlanglvl=c99complex -qlanglvl=c99complexheader

MicK7:
For Sun compiler, the -std=c99 or -xlang=c99 is required.

Maybe config.ac should have this line:
AC_PROG_CC_C99
and CMake :
set (CMAKE_C_STANDARD 99)

brtnfld:
I'll try that. You should add to the Release notes that a C99 compiler is required, starting with this release (at least I think this is the case, I've not tried it with C89).

MicK7:
I check with PGI 20.1 and it works. The older version of PGI claim to support C99 but it is not true.
I may add a configure check to enable complex.h at compile time with a MACRO.

PR #206, opened 2020-08-04
Add a warning for BC that should be of type FamilySpecified.

MicK7:
This is a small improvement patch for cgnscheck.
It advertises when a BC should be of type FamilySpecified (ie when a FamilyName is found).


brtnfld:
BTW, the \n is not needed in the string.

PR #207, opened 2020-08-06
fix cg_conn_write

MicK7:
Verify input of zone grid connectivities only when location is equal to Vertex or CellCenter

In case of FaceCenter connectivities no check is done and the library trusts the user for providing correct data.
Indeed, the library do not have the total number of faces easily available for a check while the user application should have this information.


MicK7:
This small patch should solve https://cgnsorg.atlassian.net/browse/CGNS-211

PR #208, opened 2020-08-11
some changes for vs2019 

MicK7:
cherry picking from existing patch branch


src/tools/CMakeLists.txt
```diff
@@ -5,6 +5,10 @@
 # Link all the executables to cgns and hdf5
 if (CGNS_USE_SHARED)
   link_libraries(cgns_shared)
+  if (WIN32 OR CYGWIN)
+    target_compile_definitions(cgns_shared PRIVATE -DBUILD_DLL)
+    target_compile_definitions(cgns_shared INTERFACE -DUSE_DLL)
+  endif (WIN32 OR CYGWIN)
```
brtnfld, line 11: We don't test with CYGWIN, so I'm not sure if this will work. Seems ok.

src/Test_UserGuideCode/C_code/CMakeLists.txt
```diff
@@ -3,7 +3,12 @@
 #########
 
 # Link all the executables to cgns and hdf5
-link_libraries(cgns_static)
```
brtnfld, line 6: CMakeList.txt in C_code_parallel also need this change

brtnfld:
Also, appveyor only builds SHARED=off, and on VS2015. We should add V2019 to the build matrix and enable shared builds

PR #209, opened 2020-08-13
testing updated appvoyer config.

brtnfld:
update appveyor testing to VS 19, change HDF5 binary path download


brtnfld:
@MicK7 thanks for fixing this, I was going around in circles. It's interesting that 2019 VS takes about half the time to complete.

MicK7:
@brtnfld  You are welcome.
The timing is quite an amazing fact indeed.

PR #215, opened 2020-09-22
Avoid CMake duplicate symbol error

gsjaardema:
When configuring CGNS with a CMake-build/install of HDF5, the `add_executable(h5dump IMPORTED)` command will generate the error:
```
CMake Error at CMakeLists.txt:219 (add_executable):
  add_executable cannot create imported target "h5dump" because another
  target with the same name already exists.
```
Since the file `hdf5-targets.cmake` contains the same command.  I'm not sure if this if branch is entered if HDF5 is installed via autotools, but in my testing I got a valid build with both cmake and autotools HDF5 builds.


gsjaardema:
Looks like the removed line is needed on the windows build... Will investigate further.

MicK7:
Would it be possible to remove all references to h5dump in cmake ? I can't see where it is useful.

gsjaardema:
@MicK7 Yes, that looks like it should work.  It only shows up in the top-level CMakeLists.txt as best I can see...  Everything seems to work if it is removed.

brtnfld (approved):

PR #216, opened 2020-09-24
ENABLE 64BIT by default

MicK7:
Update cmake to enable 64bit configuration by default.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -46,7 +46,7 @@ CHECK_TYPE_SIZE(long SIZE_OF_LONG)
 CHECK_TYPE_SIZE(off_t SIZE_OF_OFF_T)
 
 if ("${SIZE_OF_OFF_T}" LESS 8)
-  option(CGNS_ENABLE_LFS "Enable or disable large file support" "ON")
+  option(CGNS_ENABLE_LFS "Enable or disable large file support" "OFF")
 endif ("${SIZE_OF_OFF_T}" LESS 8)
 
 if ("${CMAKE_SIZEOF_VOID_P}" GREATER 4)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -46,11 +46,11 @@ CHECK_TYPE_SIZE(long SIZE_OF_LONG)
 CHECK_TYPE_SIZE(off_t SIZE_OF_OFF_T)
 
 if ("${SIZE_OF_OFF_T}" LESS 8)
-  option(CGNS_ENABLE_LFS "Enable or disable large file support" "OFF")
+  option(CGNS_ENABLE_LFS "Enable or disable large file support" "ON")
 endif ("${SIZE_OF_OFF_T}" LESS 8)
 
 if ("${CMAKE_SIZEOF_VOID_P}" GREATER 4)
-  option(CGNS_ENABLE_64BIT "Enable or disable 64-bit code building" "OFF")
+  option(CGNS_ENABLE_64BIT "Enable or disable 64-bit code building" "ON")
 else ("${CMAKE_SIZEOF_VOID_P}" GREATER 4)
   set(CGNS_ENABLE_64BIT "OFF")
 endif ("${CMAKE_SIZEOF_VOID_P}" GREATER 4)
--- a/src/cgns_internals.c
+++ b/src/cgns_internals.c
@@ -81,7 +81,7 @@ void *cgi_malloc(size_t cnt, size_t size)
 {
     void *buf = calloc(cnt, size);
     if (buf == NULL) {
-        cgi_error("calloc failed for %d values of size %d", cnt, size);
+        cgi_error("calloc failed for %lu values of size %lu", (unsigned long)cnt, (unsigned long)size);
         exit (1);
     }
     return buf;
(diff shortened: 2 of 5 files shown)


CMakeLists.txt
```diff
@@ -46,11 +46,11 @@ CHECK_TYPE_SIZE(long SIZE_OF_LONG)
 CHECK_TYPE_SIZE(off_t SIZE_OF_OFF_T)
 
 if ("${SIZE_OF_OFF_T}" LESS 8)
-  option(CGNS_ENABLE_LFS "Enable or disable large file support" "OFF")
+  option(CGNS_ENABLE_LFS "Enable or disable large file support" "ON")
```
gsjaardema, line 49: Is this actually related to enabling 64 bit?  Based on documentation it is for large file support which is support for files larger than 2GB which can happen with or without the 64BIT being enabled or disabled.

src/cgns_internals.c
```diff
@@ -81,7 +81,7 @@ void *cgi_malloc(size_t cnt, size_t size)
 {
     void *buf = calloc(cnt, size);
     if (buf == NULL) {
-        cgi_error("calloc failed for %d values of size %d", cnt, size);
+        cgi_error("calloc failed for %lu values of size %lu", (unsigned long)cnt, (unsigned long)size);
```
gsjaardema, line 84: I have heard, but not verified, that the size of `unsigned long` on 64-bit windows is 32 bits where the size of `size_t` is 64 bits.  If that is the case, then this can result in a truncation and meaningless error message since it would be possible for a `cnt` or `size` to be truncated in the error message.

If we are going to make this change, then there should be some compile-time assert that `sizeof(unsigned long) >= sizeof(size_t)` so we don't truncate values with these casts...

CMakeLists.txt
MicK7: You are right this is not related directly to 64bit. It is just that most of the time when 64 bit is required, LFS will be needed. In fact as the cmake is currently written, it activates only on Windows OS as far as I know.

src/cgns_internals.c
MicK7: I try to solve the problem by using %zu instead

src/cgns_internals.c
```diff
@@ -81,7 +81,7 @@ void *cgi_malloc(size_t cnt, size_t size)
 {
     void *buf = calloc(cnt, size);
     if (buf == NULL) {
-        cgi_error("calloc failed for %d values of size %d", cnt, size);
+        cgi_error("calloc failed for %zu values of size %zu", cnt, size);
```
gsjaardema, line 84: This is good as long as we use C99 (or later) compiler.

brtnfld: CGNS already requires C99+ to compile.

brtnfld:
What about changes to autotools?

brtnfld (commented):
I'm wondering if we should implement this feature before flipping the switch.

https://cgnsorg.atlassian.net/projects/CGNS/issues/CGNS-212?filter=allissues

it would be nice to allow a user to write 32Bit if they know it will fit, even if CGNS was compiled 64 bit.

CMakeLists.txt
brtnfld: Ref. https://cgnsorg.atlassian.net/browse/CGNS-184

brtnfld:
Ref. https://cgnsorg.atlassian.net/browse/CGNS-224

MicK7:
> 
> 
> I'm wondering if we should implement this feature before flipping the switch.
> 
> https://cgnsorg.atlassian.net/projects/CGNS/issues/CGNS-212?filter=allissues
>
> it would be nice to allow a user to write 32Bit if they know it will fit, even if CGNS was compiled 64 bit.

As written in the comments, the experimental_cgp_poly branch provide the feature to give the datatype to store unstructured elements in 32bit or 64bit. I don't know if this enough for a merge proposal.
The added function are : 
- CGNSDLL int cg_section_general_write(int file_number, int B, int Z,
        const char * SectionName, const CGNS_ENUMT(ElementType_t) type,
        const CGNS_ENUMT(DataType_t) elementDataType, cgsize_t start,
        cgsize_t end, cgsize_t elementDataSize, int nbndry, int *S);
- CGNSDLL int cg_elements_general_write(int fn, int B, int Z, int S,
	cgsize_t start, cgsize_t end, CGNS_ENUMT(DataType_t) m_type,
        const void *elements);
- CGNSDLL int cg_poly_elements_general_write(int fn, int B, int Z, int S,
    cgsize_t start, cgsize_t end, CGNS_ENUMT(DataType_t) m_type,
    const void *elements, const void *connect_offset);

This helps refactor lots of code parts. Other functions are helpers or experiments for parallel writing.

brtnfld:
Do you have equiv. read functions? Otherwise, there would be type conversion (in HDF5), which would be nice to avoid.

MicK7:
Equivalent read function are not implemented yet. Feedback on the writing code and API is most welcome before going forward.
Of course avoiding type conversion is nice but it is hard to avoid it completely without breaking the MLL.

MicK7:
@brtnfld after much tought, I suggest flipping the switch before implementing the feature of CGNS-212 since it will be pretty hard to get the API right for CGNS-212. Moreoverusers can still go back to previous default configuration by passing the correct CMake flags.

brtnfld:
Sounds good, but I would remove the change for LFS from this PR since that issue probably needs further review.

brtnfld (approved):
Please reference "CGNS-224" in the commit comment so the PR is linked to the issue in bitbucket.

brtnfld (commented):
And the configure needs to be updated with this change.

brtnfld (commented):
How do you build 32bit with autotools if you change it to always be true for 64, also you need to run autoconf (prefer. version 2.69) to generate configure.

MicK7:
@brtnfld To force 32 bit with autotools, one just run ./configure --enable-64bit=no

PR #221, opened 2020-11-19
Writing elements with datatype different from cgsize_t

MicK7:
-  refactor writing functions while allowing datatype different from cgsize_t (to write 32bit in a 64bit env)
 -  similarly add reading function to handle 32/64 bit elements (Not very efficient because all MLL is based upon cgsize_t and it would require to rethink a lot of code)
 -  no fortran mapping for these experimental API done or planned

see: [https://cgnsorg.atlassian.net/jira/software/c/projects/CGNS/issues/CGNS-212](https://cgnsorg.atlassian.net/jira/software/c/projects/CGNS/issues/CGNS-212)


MicK7:
@vmoschou any feedback on this proposal would be appreciated

brtnfld (commented):
Nothing obvious looks wrong, but are you planning on adding new tests?

vmoschou:
The code seems to do what we need. You have separated the memory data type (m_type) from the elementDataType and this is our case.

The m_type is defined in _cg_elements_general_write_ ( _cg_poly_elements_general_write_ ) and  and the _elementDataType_ is defined in _cg_section_general_write_.

Our _m_type_ is _int64_t_ and we need the _elementDataType_ to be _int32_t_.

Input functions can do the tranformation as well.

So I believe we are fine.

MicK7:
@vmoschou Thanks for the feedback
@brtnfld  Thanks for the review. Most of the new code is called by existing test code due to refactoring. But I plan to add a parallel test and a serial test to cover parts that need to be tested.

brtnfld:
I created subtasks for tests and documentation (and Fortran APIs). Is there a reason why this can't be released in 4.2? It is targeted for the 5.0 release.

MicK7:
@brtnfld I don't know why it could not be released in 4.2

brtnfld (approved):
please include the JIRA issue in the commit log.

PR #222, opened 2020-12-10
update osx

MicK7:
Apple and Homebrew no longer support MacOS 10.13. This means homebrew no longer provides binaries for MacOS 10.13, and on those systems, everything has to be built from source. This is leading to a timeout in travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
-osx_image: xcode9.4
+osx_image: xcode12
 dist: xenial
 language: C
 sudo: required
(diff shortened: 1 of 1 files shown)


brtnfld (approved):

brtnfld:
We should remove the OSX build and find an alternative:


https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing

What do I do if I can’t build my builds?

After November 1st, all users on updated concurrency plans who wish to run macOS builds will need to go to the plan page in your account settings and purchase credit addons in order to run your builds.

Additionally, all users who are on the new usage based plan (previously on the 10 concurrent job plan and higher), will start with a credit allotment but will have to purchase credit addons if you run out of credits.

MicK7:
Thanks for the link.
It seems that using travis is becoming harder.

According to this blog post [travis-ci threw wrench my open source works](https://www.jeffgeerling.com/blog/2020/travis-cis-new-pricing-plan-threw-wrench-my-open-source-works) , Github Actions could be a valid alternative. However, it provides only 2000 min per month and thus only push to the main branch should be taken into account.

PR #224, opened 2020-12-11
fix MPI_Recv in test_poly_unstructured

MicK7:


MicK7:
@brtnfld This should definitely close the issue due to bad type passed to MPI. So sorry.

brtnfld:
Isn't there still an issue with the test needed to be run with 2 ranks?

MicK7:
The test needed to be run on 2 ranks is just a Warning  (it comes from a copy of the warning in test_general_readwrite) but it is safe to run the test for a larger number. (I did tests with 8 ranks).

brtnfld (approved):

PR #226, opened 2021-01-28
CGNS-231 test_convert_elem fails with gcc 10

brtnfld:
(1) Fixed "size" scope with LongInteger
(2) I think it is also an error if size=0
(3) checked array state before trying to free it.


src/cgnslib.c
```diff
@@ -5039,12 +5039,13 @@ int cg_poly_elements_general_read(int file_number, int B, int Z, int S,
     else if (m_type == CGNS_ENUMV(LongInteger)) {
         cglong_t* tmp_connect_offset = (cglong_t*)connect_offset;
         cglong_t offset = tmp_connect_offset[0];
-        cglong_t size = tmp_connect_offset[end - start + 1] - offset;
-        if (size < 0) return CG_ERROR;
+        cglong_t size_long = tmp_connect_offset[end - start + 1] - offset;
+        if (size_long < 1) return CG_ERROR;
+        size = (cgsize_t)size_long;
```
brtnfld, line 5044: Shouldn't we check if this will fit before assigning it?

MicK7: Yes, it would be safer to add a check to limit overflow.
Right now, since "size_long" is the result of a difference the problem would happen only if the user try to read a big array at once and not partially.

MicK7 (approved):

src/cgnslib.c
```diff
@@ -5084,12 +5085,12 @@ int cg_poly_elements_general_read(int file_number, int B, int Z, int S,
             s_start, s_end, s_stride,
             section->connect->data_type,
             1, m_dim, m_start, m_end, m_stride, conv_data)) {
-            free(conv_data);
+            if (conv_data) free(conv_data);
```
MicK7, line 5088: why do we need to check conv_data ?
I assume it was ok because of line 5080

brtnfld: fixed, I also removed OSX CI testing.

PR #228, opened 2021-02-26
replaced the use of H5Literate_by_name with H5Literate

brtnfld:
H5Literate_by_name is not regression tested in HDF5, H5Literate is.


MicK7:
Looks good to me. Does it help in closing CGNS-194 ?

MicK7 (approved):

brtnfld:
No, H5Literate_by_name is ok supported for non-POSIX systems.

PR #231, opened 2021-03-23
Fix for CGNS-240

brtnfld:
CGNS-240

Made CG_MODE_CLOSED private, removed from public headers and Fortran constants.


MicK7 (approved):
Indeed using CG_MODE_CLOSE with cg_open leads to an error

PR #232, opened 2021-03-24
CGNS-241, fixed cg_configure return value error

brtnfld:
Fixes false error returned from cg_configure and added error checks for cg_configure in the tests, CGNS-241.

### DIFF
--- a/src/cgnslib.c
+++ b/src/cgnslib.c
@@ -773,7 +773,7 @@ int cg_configure(int what, void *value)
 {
     /* cgio options */
     if (what > 100) {
-      if( cgio_configure(what, value) != -1) {
+      if( cgio_configure(what, value) != CG_OK) {
         cg_io_error("cgio_configure");
         return CG_ERROR;
       }
--- a/src/cgnslib.c
+++ b/src/cgnslib.c
@@ -773,10 +773,10 @@ int cg_configure(int what, void *value)
 {
     /* cgio options */
     if (what > 100) {
-        if (cgio_configure(what, value)) {
-            cg_io_error("cgio_configure");
-            return CG_ERROR;
-        }
+      if( cgio_configure(what, value) != -1) {
+        cg_io_error("cgio_configure");
+        return CG_ERROR;
+      }
     }
     /* error message handler */
     else if (what == CG_CONFIG_ERROR) {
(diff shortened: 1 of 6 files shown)


src/cgnslib.c
```diff
@@ -773,10 +773,10 @@ int cg_configure(int what, void *value)
 {
     /* cgio options */
     if (what > 100) {
-        if (cgio_configure(what, value)) {
-            cg_io_error("cgio_configure");
-            return CG_ERROR;
-        }
+      if( cgio_configure(what, value) != -1) {
```
MicK7, line 776: Hard coded error '-1' comes from the NO_ERROR def of adfh.h that is unreachable at cgnslib level.

Thus, the issue comes from CGIO_ERR_BAD_OPTION == NO_ERROR (from adfh.h)
It would be better to fix the problem at cgio level.

In cgnslib.c we should have something like :
if (cgio_configure:(what, value) != CGIO_NO_ERROR) {

while in cgns_io.c line 543, a modification could be:
#if CG_BUILD_HDF5
        ADFH_Configure(what-200, value, &ierr);
        if (ierr == NO_ERROR) {
           ierr = CGIO_NO_ERROR;
        }
#endif

It would keep the CGIO_ERR_NONE valid for users who use cgns at cgio level and
let propagate hdf error code for other cases.

An other more simple solution is just to change adfh.h at line 55:
#define NO_ERROR		       -1 --> 0
... to be coherent with cgio and hdf5...

In the end, I don't really like using hardcoded value to check for error.

MicK7 (changes requested):

src/cgnslib.c
brtnfld: Good call. I changed NO_ERROR to 0 instead and updated the code to reflect this change.

I'm not sure that cgio_configure should be a public function because, as is, it only has meaningful context with the cg APIs. And it gives the impression that it is similar to cg_configure. But some parameters can be set with cg_configure and not cgio_configure. I don't know the history of cgio_configure, so I don't know why it was designed as a public function, and it is not documented in the user's guide, which is an issue for a public function.

MicK7: cgio_configure is meant to configure the database layer (hdf5 mainly) while cg_configure define also some global behavior of the mll top layer (cpex behavior for rind planes, ... ; error_handler).
Since I have seen some code using directly the cgio layer, I think that keeping cgio_configure public is good.
Thus having an issue on the bad documentation is alright and need to be addressed.

MicK7: IMHO, we could even imagine to use cg_configure to force reading file of newer cgns library file with an old library.

MicK7 (approved):

PR #238, opened 2021-04-02
small tuning

MicK7:
use memcpy instead of strcpy and skip costly string copy when name will not move


MicK7:
@brtnfld This should be closer to original implementation for small number of zone.

brtnfld (approved):
looks good

PR #243, opened 2021-04-08
fix fortran failure

MicK7:
With some fortran compiler, cgwrite_f03 is failing.
Indeed the fortran to C mapping does not go well and a (void **) was transfered thus value_c did not get the right value (the set value in value_c was the pointer address). By explicitly casting the whole chain, it solves the issue.

Furthermore value_f is declared as "cgsize_t" while "int" is used inside the C code. (In particular, we have ((int)(size_t)file_type) ...
Changing between 64bit and 32bit compilation generates also a failed test.

### DIFF
--- a/src/cg_ftoc.c
+++ b/src/cg_ftoc.c
@@ -180,8 +180,7 @@ CGNSDLL void cg_set_file_type_f(cgint_f *ft, cgint_f *ier)
 
 /*-----------------------------------------------------------------------*/
 
-CGNSDLL void cg_configure_f(cgint_f *what,
-        void *value, cgint_f *ier)
+CGNSDLL void cg_configure_f(cgint_f *what, void *value, cgint_f *ier)
 {
   size_t value_c;
 
--- a/src/cg_ftoc.c
+++ b/src/cg_ftoc.c
@@ -180,10 +180,8 @@ CGNSDLL void cg_set_file_type_f(cgint_f *ft, cgint_f *ier)
 
 /*-----------------------------------------------------------------------*/
 
-CGNSDLL void cg_configure_c(cgint_f *what, void *value, cgint_f *ier)
+CGNSDLL void cg_configure_c_ptr(cgint_f *what, void *value, cgint_f *ier)
 {
-  size_t value_c;
-
   /* CHARACTERS */
   if( (int)*what == CG_CONFIG_SET_PATH ||
       (int)*what == CG_CONFIG_ADD_PATH) {
@@ -208,13 +206,21 @@ CGNSDLL void cg_configure_c(cgint_f *what, void *value, cgint_f *ier)
       return;
     }
 
-  /* FUNCTION POINTER */  
-  } else if( (int)*what == CG_CONFIG_ERROR ) {
-    *ier = (cgint_f)cg_configure((int)*what, value);
-
   /* EVERYTHING ELSE */
   } else {
     *ier = (cgint_f)cg_configure((int)*what, (void *)(*(size_t *)value));
+  }
+}
+/*-----------------------------------------------------------------------*/
+
+CGNSDLL void cg_configure_c_funptr(cgint_f *what, void *value, cgint_f *ier)
+{
+  /* FUNCTION POINTER */
+  printf("lkdjsflk");
+  if( (int)*what == CG_CONFIG_ERROR ) {
+    *ier = (cgint_f)cg_configure((int)*what, value);
+  } else {
+    *ier = (cgint_f)CG_ERROR;
   }
 }
 
--- a/src/cg_ftoc.c
+++ b/src/cg_ftoc.c
@@ -180,22 +180,40 @@ CGNSDLL void cg_set_file_type_f(cgint_f *ft, cgint_f *ier)
 
 /*-----------------------------------------------------------------------*/
 
-CGNSDLL void FMNAME(cg_configure_f, CG_CONFIGURE_F) (cgint_f *what,
+CGNSDLL void cg_configure_f(cgint_f *what,
         void *value, cgint_f *ier)
 {
   size_t value_c;
-  if( (int)*what != CG_CONFIG_ERROR &&
-      (int)*what != CG_CONFIG_SET_PATH &&
-      (int)*what != CG_CONFIG_ADD_PATH &&
-      (int)*what != CG_CONFIG_HDF5_MPI_COMM) {
-    value_c = *((int *) *((size_t *)value));
-    *ier = (cgint_f)cg_configure((int)*what, (void*)value_c);
+
+  /* CHARACTERS */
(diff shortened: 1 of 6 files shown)


brtnfld (commented):
Also,FMNAME(cg_configure_f, CG_CONFIGURE_F) needs to be changed back because BIND(C,NAME='cg_configure_f') was used. What was the intel warning with this?

src/cg_ftoc.c
```diff
@@ -188,7 +188,7 @@ CGNSDLL void FMNAME(cg_configure_f, CG_CONFIGURE_F) (cgint_f *what,
       (int)*what != CG_CONFIG_SET_PATH &&
       (int)*what != CG_CONFIG_ADD_PATH &&
       (int)*what != CG_CONFIG_HDF5_MPI_COMM) {
-    value_c = *(size_t *)value;
+    value_c = *((int *) *((size_t *)value));
```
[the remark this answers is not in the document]
brtnfld: 191 also segfaults with gcc/gfortran 7.5 at the first call to cg_configure_f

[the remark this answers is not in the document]
MicK7: cg_configure_f binding was not found. 
After adding the FMNAME it was working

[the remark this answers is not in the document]
brtnfld: I missed adding it to cgns_fortrandll.def.in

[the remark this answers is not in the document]
MicK7: Yes HDF5 option that are of "size_t" sizes need a separate statement.
Beware that cgsize_t != size_t 
cgsize_t depends on CGNS compilation option while size_t depends on external stdlib.h .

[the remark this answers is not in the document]
brtnfld: It looks like (void *)(*(size_t *)value works for everything except CG_CONFIG_RIND_INDEX.

src/cg_ftoc.c
```diff
@@ -197,7 +197,7 @@ CGNSDLL void FMNAME(cg_configure_f, CG_CONFIGURE_F) (cgint_f *what,
     *ier = (cgint_f)cg_configure((int)*what, &C_comm);
 #endif
   } else {
-    *ier = (cgint_f)cg_configure((int)*what, value);
+    *ier = (cgint_f)cg_configure((int)*what, (void *)(*(size_t *)value));
```
brtnfld, line 200: This change segfaults

MicK7: When does it segfaults ? with which option ?

brtnfld: #0  0x000000000043ea20 in cg_set_path (path=0x2f612f747365742f <error: Cannot access memory at address 0x2f612f747365742f>) at cgnslib.c:839
#1  0x0000000000406f3c in cg_configure_f (what=what@entry=0x8cf3c4, value=value@entry=0x7fffffffaf80, ier=ier@entry=0x7fffffffac24) at cg_ftoc.c:209
#2  0x00000000004058ef in write_cgns_1 () at cgwrite_f03.F90:359
#3  0x000000000040424f in main (argc=<optimized out>, argv=<optimized out>) at cgwrite_f03.F90:32
#4  0x000015555388534a in __libc_start_main () from /lib64/libc.so.6
#5  0x000000000040428a in _start () at ../sysdeps/x86_64/start.S:120

/configure --verbose \
--enable-debug=-g \
--enable-parallel \
--with-fortran \
--with-hdf5="$HDF5" \
--enable-lfs \
--enable-cgnstools \
--disable-shared \

brtnfld: Was it the intel compiler that had an issue with this line originally?

MicK7: Yes, it was generating this issue when reaching the function callback setting :
``
#0  0x00000000006b8aa8 in write_cgns_1_$FUNC.0.3 ()
#1  0x00007ffff7ac31c7 in cgi_error (format=0x1 <Address 0x1 out of bounds>)
    at /home/mickael/CGNS/src/cgns_error.c:38
#2  0x00007ffff7b02c6f in cg_set_file_type (file_type=1) at /home/mickael/CGNS/src/cgnslib.c:733
#3  0x00007ffff7b60108 in cg_configure_f_ (what=0x1, value=0x7ffff7db99c0 <cgns_error_mess>, 
    ier=0x7ffff7db99e2 <cgns_error_mess+34>) at /home/mickael/CGNS/src/cg_ftoc.c:192
#4  0x000000000040bcc1 in write_cgns_1 () at /home/mickael/CGNS/src/tests/cgwrite_f03.F90:373
#5  0x000000000040965e in main ()
#6  0x00007ffff5962c05 in __libc_start_main () from /lib64/libc.so.6
#7  0x0000000000409569 in _start ()
``

MicK7: (void *)(*(size_t *)value

was needed to pass properly the c_funloc

MicK7: Maybe separating CG_CONFIG_ERROR is needed

brtnfld: Yes, looks like it.

brtnfld: are you working on a push, or should I do it for what I have so far?

MicK7: You can do it with what you have so far.
I have not found any improvement to push right now.

src/cg_ftoc.c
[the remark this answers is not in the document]
brtnfld: For HDF5 configure args that are size_t, I think we should require the type to be C_SIZE_T. That way we know it will be compatible with the HDF5 APIs. It seems to make sense because they are configuring HDF5 directly and bypassing CGNS.

MicK7:
In cgns_f.F90, intel fortran complains about c_loc and c_funloc mismatch.
Adding the following directive helps compiling:

```
     TYPE(C_PTR), VALUE :: value
!DIR$ ATTRIBUTES NO_ARG_CHECK :: value
     INTEGER, INTENT(OUT) :: ier
```

brtnfld:
Ah yes, I see the issue. I need to overload cg_configure_f with TYPE(C_FUNPTR) instead of TYPE(C_PTR).

MicK7:
cmake is not mangling correctly the name cg_configure_f . I still have not found where is the difference with autotools or if it is my cmake install that is buggy...

MicK7:
Current branch works well with cmake and autotools with intel and recent gnu compiler.
Looks ready for merging

brtnfld:
I'm just waiting for our tests to run. 
There does some to be a bug in gfortran with picking the correct interface if the example uses a variable instead of in-line. I've submitted a bug report, https://gcc.gnu.org/bugzilla/show_bug.cgi?id=99982

MicK7:
Nice catch

brtnfld (approved):

PR #244, opened 2021-04-14
Fixed to many communicators, CGNS-109

brtnfld:
Leaving HDF5 fapl ids open caused MPI_Comm errors when repeatedly opening and closing a CGNS file.
CGNS now closes the fapl. Added test which opens and closes file repeatedly.


MicK7 (approved):
LGTM
Tested with intelmpi

PR #245, opened 2021-04-15
CGNS-141 fixed failures in mixing serial parallel file creations

brtnfld:
Excluded parallel HDF5 APIs from non-MPI IO FD usage, CGNS-141


MicK7 (approved):

MicK7:
When I see the externs that are pilling up in adfh I wonder if adding a struct like the following in cgns_io would be worth if:

struct {
  char hdf5_access[64] = "NATIVE";
#if CG_BUILD_PARALLEL
  int pcg_mpi_comm_size;
  int pcg_mpi_comm_rank;
  int pcg_mpi_initialized;
  MPI_Info pcg_mpi_info;
#endif
} cgio_ctxt;

Only having one external global context would be nice.

brtnfld:
That would be cleaner. I'll work on changing it.

MicK7:
@brtnfld I suggest to merge as it is and create a new jira issue of improvement for the next release.

MicK7:
I tested adding a context struct  https://github.com/MicK7/CGNS/commit/3592a3fdf6a1409338809c662172ae2a7a12d3e4 
The struct is not including default_pio_mode but maybe it should.

brtnfld:
@MicK7 sorry, I merged it without addressing default_pio_mode; it should probably be added to the struct.

PR #246, opened 2021-04-16
CGNS-233 Does configure.ac need to check for zlib/szip headers?

MicK7:
Remove header checking for zlib and szip.

### DIFF
--- a/src/configure
+++ b/src/configure
@@ -6128,136 +6128,41 @@ else
 fi
 
 
-  case $withval in
-    yes)
-      H5NEEDZLIB=1
-      HAVE_ZLIB="yes"
-      for ac_header in zlib.h
-do :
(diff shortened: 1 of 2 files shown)


MicK7:
Regenerating with autoconf the configure file is needed before merging.

MicK7:
@markcmiller86   Thanks for reporting the autotools CGNS issue. Here is a quick hack of the CGNS configure that hopes to be included in next release. Do you think it would be enough for VisIt ?

markcmiller86:
LGTM! Thanks so much!

brtnfld:
can you run Autoconf and commit configure

MicK7:
@brtnfld  Sorry I don't have access to any unix machine right now and will not have access to one next week too.
If you could add the configure I would be thankful.

MicK7:
@brtnfld Thanks so much

brtnfld (approved):

PR #260, opened 2021-04-26
Synchronize develop with master

MicK7:


brtnfld (commented):
I would only merge HISTORY.txt, RELEASE.txt is a blank template to be filled out when released.

MicK7:
Ok sorry I will drop the request then.
I was thinking of having the two branch identical after release to have a cleaner reading of future merge but it is not mandatory.

brtnfld:
HISTORY.txt usually gets changed with the PR to change the version number of develop.

PR #262, opened 2021-04-28
CGNS-247

MicK7:
Add a context structure for cgio and hdf5 related state

### DIFF
--- a/src/pcgnslib.c
+++ b/src/pcgnslib.c
@@ -35,15 +35,19 @@ freely, subject to the following restrictions:
                              (type >= CGNS_ENUMV(BAR_4) && \
                               type <= CGNS_ENUMV(HEXA_125)))
 
-/* MPI-2 info object */
-extern MPI_Info pcg_mpi_info;
-extern MPI_Comm pcg_mpi_comm;
-extern int pcg_mpi_comm_size;
-extern int pcg_mpi_comm_rank;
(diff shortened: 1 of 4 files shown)


src/pcgnslib.c
```diff
@@ -35,17 +35,20 @@ freely, subject to the following restrictions:
...
-extern MPI_Comm pcg_mpi_comm;
-extern int pcg_mpi_comm_size;
-extern int pcg_mpi_comm_rank;
-/* Flag indicating if HDF5 file accesses is PARALLEL or NATIVE */
-extern char hdf5_access[64];
-/* flag indicating if mpi_initialized was called */
-extern int pcg_mpi_initialized;
-
-hid_t default_pio_mode = H5FD_MPIO_COLLECTIVE;
+typedef struct _cgns_io_ctx_t {
```
gsjaardema, line 38: It looks like we have the same typedef defined both here and in `src/cgns_io.c`.  Shouldn't it be defined once in an include file and included both places?

MicK7: Good idea. It should be better in an header file but the header should be private not to expose this structure outside of cgns.

gsjaardema: Looks much cleaner.

gsjaardema (approved):
Looks good to me.  Much cleaner overall

brtnfld (approved):
LGTM

PR #263, opened 2021-04-30
cgi_read_base pattern modification

MicK7:
When opening a CGNS file with MLL the whole file is parsed. Moreover the zone are sorted alphabetically.
This little read pattern modification propose to use qsort to sort the zone by name.
It also add a small "_children_t vector" to cache node data read during base node parsing. Thus it should prevent multiple small reads.


src/cgns_internals.c
```diff
@@ -121,6 +124,201 @@ int cgi_read()
     return CG_OK;
 }
 
+/* helpers to parse CGNSBase_t node */
+typedef enum {
```
MicK7, line 128: An enum makes it easier to compare than a string for later filtering

src/cgns_internals.c
```diff
@@ -121,6 +124,201 @@ int cgi_read()
...
+    }
+    else if (0 == strcmp(nodelabel, "UserDefinedData_t")) {
+        return LabelUserDefinedData_t;
+    }
+    else {
+        return LabelNull_t;
+    }
+}
+
+typedef struct _childnode {
```
MicK7, line 198: Not sure if it a good idea to include the name

src/cgns_internals.c
```diff
@@ -121,6 +124,201 @@ int cgi_read()
...
+    return (strcmp(p1->name, p2->name));
+}
+
+#ifdef _WIN32
+#define CG_INLINE __inline
+#else
+#define CG_INLINE static inline
+#endif
+
+CG_INLINE int cgi_get_nodes_with_label(const _childnode_t *childlist, const int nchildren, BaseLabel_t label, int *nnodes, int *startpos)
```
MicK7, line 230: This return a section of the whold childlist that starts at starpos and ends at starpos+nnodes
Maybe returning a new array with only the node would be as efficient and remove the need of a sort by label

src/cgns_internals.c
```diff
@@ -187,73 +387,80 @@ int cgi_read_base(cgns_base *base)
...
-    if (cgi_get_nodes(base->id, "Zone_t", &base->nzones, &id)) return CG_ERROR;
+    cgi_get_nodes_with_label(childlist, nchildren, LabelZone_t, &base->nzones, &start);
     if (base->nzones>0) {
          /* Order zones alpha-numerically */
-        if (cgi_sort_names(base->nzones, id)) {
-            cgi_error("Error sorting zone names...");
-            return CG_ERROR;
-        }
-         /* read & save zones in sorted order */
+        mergesort(&childlist[start], base->nzones, sizeof(_childnode_t), sort_zone_names);
```
MicK7, line 450: Here this is supposed to be more efficient than original code
It is only here that the names are really required.
Maybe use qsort that is official C to not include merge sort would be more sustainable

src/cgns_internals.c
```diff
@@ -187,73 +387,80 @@ int cgi_read_base(cgns_base *base)
...
+         /* populate zones in sorted order */
         base->zone = CGNS_NEW(cgns_zone, base->nzones);
         for (n=0; n<base->nzones; n++) {
-            base->zone[n].id = id[n];
-            base->zone[n].link = cgi_read_link(id[n]);
+            base->zone[n].id = childlist[n+start].id;
+            base->zone[n].link = cgi_read_link(childlist[n + start].id);
             base->zone[n].in_link = 0;
-            if (cgi_read_zone(&base->zone[n])) return CG_ERROR;
+            strcpy(base->zone[n].name, childlist[n + start].name);
```
MicK7, line 457: Do not read again the node name

src/cgns_internals.c
```diff
@@ -187,73 +387,80 @@ int cgi_read_base(cgns_base *base)
...
         }
-        CGNS_FREE(id);
+    }
+    CGNS_FREE(childlist);
+    /* read zones */
+    for (n = 0; n < base->nzones; n++) {
+        if (cgi_read_zone(&base->zone[n])) return CG_ERROR;
```
MicK7, line 463: Read each zone but do not read again the zone name

src/cgns_internals.c
```diff
@@ -267,9 +474,8 @@ int cgi_read_zone(cgns_zone *zone)
     double *id;
     cgsize_t dim_vals[12];
 
-     /* Zone_t */
-    if (cgi_read_node(zone->id, zone->name, data_type, &ndim, dim_vals,
-            &vdata, READ_DATA)) {
+     /* Zone_t, assume zone->name is already read */
+    if (cgi_read_node_data(zone->id, data_type, &ndim, dim_vals, &vdata)) {
```
MicK7, line 478: Only read the node data of the Zone_t and not its name

src/cgns_internals.c
```diff
@@ -121,6 +124,201 @@ int cgi_read()
...
+        *childnodes = childlist;
+        *nnodes = nid;
+    }
+    else {
+        CGNS_FREE(childlist);
+    }
+    return CG_OK;
+}
+
+int cgi_read_equations_from_list(int in_link, _childnode_t* nodelist, int nnodes, cgns_equations** equations);
```
MicK7, line 309: cgi_read_*_list are private helper function that are only valid for reading CGNSBase_t node
The original function they are derived from are kept since they are possibly called by other part of the code or cgnslib.c

MicK7:
@brtnfld Here are some coding thoughts to improve the situation. If it can help get started with fixing CGNS-220.

One of the question I am still wondering is : how can we not load the Zone_t structure on all the MPI process and store the collection of Zones in a distributed way so that not a lot of memory is used per processor ? The CGNS/MLL API seems quite limited in this case and designing a new API directly for HPC would be better than adapting the MLL.

brtnfld:
> @brtnfld Here are some coding thoughts to improve the situation. If it can help get started with fixing CGNS-220.
> 
> One of the question I am still wondering is: how can we not load the Zone_t structure on all the MPI process and store the collection of Zones in a distributed way so that not a lot of memory is used per processor? The CGNS/MLL API seems quite limited in this case and designing a new API directly for HPC would be better than adapting the MLL.

Storing the information in memory should help performance; it is similar to using the HDF5 core file driver. The obvious issue is coming up with a schema for how the information gets coordinated among the processes. Agreed that a new pcg API would be better.  

We would have to make sure that the in-memory reflects what is persisted to disk, and avoid reading/writing small datasets to disk.

MicK7:
I fully agree with your point of view. I was just stating that sometimes one MPI process just need to know that a Zone_t exists but will never need to access its data. Thus this process can delegate to other MPI process the caching of the data. When it really access to the Zone_t sub-nodes it could retrieve it from the other process.

MicK7:
@brtnfld Do you think these changes  could be good enough for main branch ?

brtnfld:
Is this ready for review?

MicK7:
It should be ready for review

src/cgns_internals.c
```diff
@@ -121,14 +122,220 @@ int cgi_read()
...
+    }
+}
+
+typedef struct _childnode {
+    double id;
+    int type;
+    char_33 name;
+} _childnode_t;
+
+static int sort_base_children(const void* v1, const void* v2)
```
gsjaardema, line 204: Is this used anywhere?

src/cgns_internals.c
```diff
@@ -121,14 +122,220 @@ int cgi_read()
...
+    p2 = (_childnode_t *) v2;
+    return p1->type - p2->type;
+}
+
+static int sort_zone_names(const void* v1, const void* v2)
+{
+    _childnode_t* p1;
+    _childnode_t* p2;
+
+    p1 = (_childnode_t*)v1;
```
gsjaardema, line 219: Not sure if using C99 or greater, but maybe do:
```
_childnode_t* p1 = (_childnode_t*) p1;
_childnode_t* p2 = (_childnode_t*) p2;
```
If not, then consider:
```
_childnode_t* p1 = NULL;
_childnode_t *p2 = NULL;
```

src/cgns_internals.c
```diff
@@ -121,14 +122,220 @@ int cgi_read()
...
+
+    for (n = 0; n < nchildren; n++) {
+        if (childlist[n].type != label && first) continue;
+        if (childlist[n].type != label) break;
+        if (first) {
+            start = n;
+            first = 0;
+        }
+    }
+    *nnodes = n - start;
```
gsjaardema, line 244: Does this work correctly if there is not match at all?  Should there be a test if `first == 1` after the for loop and if so, return an error?

src/cgns_internals.c
```diff
@@ -187,73 +394,186 @@ int cgi_read_base(cgns_base *base)
...
+                }
+                return CG_ERROR;
+            }
         }
-        CGNS_FREE(id);
     }
 
-     /* ReferenceState_t */
-    if (cgi_read_state(0, base->id, &base->state)) return CG_ERROR;
+    /* ReferenceState_t */
```
gsjaardema, line 436: It seems like maybe all of these could call a common function instead of repeating the body of the `if` statement for each...   Not that much code, but it is repeated eleven times...

src/cgns_internals.c
```diff
@@ -4530,26 +5463,15 @@ int cgi_read_exponents(int in_link, double parent_id, cgns_exponent **exponents)
...
-    }
-    units[0] = CGNS_NEW(cgns_units, 1);
-    units[0]->id = id[0];
-    units[0]->link = cgi_read_link(id[0]);
-    units[0]->in_link = in_link;
-    CGNS_FREE(id);
-
     if (cgi_read_string(units[0]->id, units[0]->name, &string_data)) return CG_ERROR;
-    if (strlen(string_data) != 32*5) {
+    if (strlen(string_data) != 32 * 5) {
```
gsjaardema, line 5474: Maybe replace the `32` magic number with some meaningful name...  Maybe that is a separate PR later on to not clutter this one...

gsjaardema (commented):
Looks good to me although I didn't look at everything in detail.  Did quick scan over most with some random deeper looks...

src/cgns_internals.c
MicK7: I have wrapped the "if" in a macro to make it cleaner.

src/cgns_internals.c
MicK7: The function was returning a start equals to the end of the list when there is no match at all.
The function is not needed anymore.

src/cgns_internals.c
MicK7: done

src/cgns_internals.c
MicK7: not used anymore. I removed it.

src/cgns_internals.c
MicK7: I haven't investigated from where comes this magic number. You are right a separate PR would be needed.

MicK7:
This should solve the N^2 sorting of zonename of https://cgnsorg.atlassian.net/browse/CGNS-220
I would be nice to do a benchmark with the new code

brtnfld:
I'm assuming  Metadata_Stressing.c would show the improvement, I'll benchmark it.

brtnfld:
cgns_internals.c:375:13: warning: comparison of array 'nchildbylabel' equal to a null pointer is always false [-Wtautological-pointer-compare]
        if (nchildbylabel == 0) continue;
            ^~~~~~~~~~~~~    ~

MicK7:
> cgns_internals.c:375:13: warning: comparison of array 'nchildbylabel' equal to a null pointer is always false [-Wtautological-pointer-compare]
> if (nchildbylabel == 0) continue;
> ^~~~~~~~~~~~~ ~

Thanks for catching this bug. It was supposed to be nchildbylabel[n] !!

MicK7:
@brtnfld Did you manage to benchmark after the fix ?

MicK7:
Merging even if no benchmark done right now.

brtnfld:
@MicK7 I had an error in my batch script, I'm rerunning it now.

brtnfld:
@MicK7  On Summit (ORNL), I tested Metadata_Stressing using 7056 ranks and ran it four times to get the mean, min, max over the four runs. All the timing numbers are summarized in https://cgnsorg.atlassian.net/browse/CGNS-220

The cgp_open shows an improvement from 191s to 186s (mean).

MicK7:
Thanks for testing. The improvement is not so big but very happy that it goes in the right direction. I am wondering which read timing with cg_open could be achievable without zone name sorting.

PR #266, opened 2021-05-03
struct fixes for parallel.

brtnfld:
Fixed missing struct contexts and updated needed include in make because of new struct.


MicK7 (approved):

MicK7:
Sorry for missing those lines

PR #267, opened 2021-05-05
Fixed Scoping issues (mainly with Fortran)

brtnfld:
Went down the scoping rabbit hole. Fixed all the missing scope enumerations in the Fortran tests, which lead to getting rid of the fixed formatting, Reworked CONCATENATE to work with gfortran (see cgnstypes_f03.h.in), CGNS-249

Not Fun, thankfully we have _find_ and _Perl._


MicK7:
I am not really familiar with the scoping feature so I trust others to review all the holes.
Thanks for the not so fun dig up

MicK7:
@gsjaardema @brtnfld  Do you think this will require a minor release 4.2.1 ?

brtnfld:
I would not think a minor release is needed since PR#265 fixed a test, and this fixes a previous bug/tests already in 4.1.2 (and 4.2.0); the CGNS library itself is not affected.

brtnfld:
@gsjaardema, are you going to have time to review this, or should I just merge it?

src/Test_UserGuideCode/Fortran_code/write_con2zn_str.F90
```diff
@@ -1,6 +1,7 @@
       program write_con2zn_str
       use cgns
       implicit none
+#include "cgnstypes_f03.h"
```
gsjaardema, line 4: Doesn't look ilke this one is needed.

gsjaardema (approved):
Looks good.  Tried to skim every file and everything looked like a reasonable change.  

The `cgns_io.c`file that was changed for the `mpi_initialized` looks out-of-place in these changes.  But it looks like a needed change.

brtnfld:
I don't know why the mpi_initialized change is showing up, that change is already in develop.

PR #269, opened 2021-05-12
CGNS-232, replace TravisCI with Github Actions

brtnfld:
I moved to build HDF5 and make checks from the bash scripts to the yml file instead since they are trivial.

The yml file still calls a bash script for configuring Autotools builds and CMake builds, I have the feeling that this would really complicate the yml file since bash seems to be more versatile than yml.

I fixed a few minor issues with the tests which caused Actions to fail.

I think current coverage is fairly comprehensive, but we can add more/tweak if needed, it a least gets Linux and macOS testing back for PRs once again.

And yes, some trial and error was involved in getting it to work :)

CGNS-232


MicK7:
@brtnfld Nice job !


May be adding a cmake-ninja build that should run faster than good old make, can also be considered in the future.

`cmake -G Ninja ...`
`ninja build`
`ninja test`

and dependency will be
- macOS:
`brew install ninja`
- Ubuntu
`sudo apt-get install ninja-build`

Since it seems quite easy to add, I think the feature can be thought over later.

MicK7 (approved):

PR #270, opened 2021-05-20
CGNS 244 replaced #ifdef JFC_PATCH_2015_2 with cg_configure

brtnfld:
Users can dynamically set the parameters in the old #def patch.

Added a reset configure option to go back to the default values instead of explicitly resetting the defaults.


MicK7 (approved):

src/adfh/ADFH.c
```diff
@@ -1464,14 +1480,32 @@ void ADFH_Configure(const int option, const void *value, int *err)
...
+    }
 #if CG_BUILD_PARALLEL
     else if (option == ADFH_CONFIG_MPI_COMM) {
       MPI_Comm* comm = (MPI_Comm*)value;
       if (!comm) {
         set_error(ADFH_ERR_INVALID_USER_DATA, err);
       }
       else {
-        ParallelMPICommunicator = (MPI_Comm)*comm;
+        ctx_cgio.pcg_mpi_comm = (MPI_Comm)*comm;
```
MicK7, line 1508: Oh ! nice catch

MicK7:
LGTM.

PR #271, opened 2021-06-02
Keep default library name on mingw

xantares:
For windows platforms we only need to rename the shared library when the import library name is the same as the static name
With msvc this is usually the case (cgns.lib) but not on mingw where libcgns.a and libcgns.dll.a can coexist.


xantares:
@brtnfld not really a bug, but it helps projects that depend on cgns finding the right library name

what's the status on this ?

MicK7:
I think that we need to check how it impacts existing compilation toolchain but it seems to be clearly an improvement.

brtnfld:
@MicK7, the issue I'm not clear about is taking something that was windows specific and now applying that to macOS and Linux. I don't see any issues, but I'm not 100% sure.

xantares:
it doesnt change anything on osx/linux because shared and static have different extensions, only on mingw

brtnfld (approved):
I don't see any issues with the changes.

PR #284, opened 2021-06-30
WIP: [DO NOT MERGE] feature -- HDF5 chunked datasets and compression

brtnfld:
Initial implementation for adding HDF5 chunked dataset option and HDF5 filter (compression) option.


src/cgnslib.c
```diff
@@ -833,6 +833,14 @@ int cg_get_compress(int *compress)
     return CG_OK;
 }
 
+int cg_set_filter(const struct CG_FILTER *filter)
```
brtnfld, line 836: @MicK7 Do you think it is ok to use a struct to pass the filter API parameters? It does not look like it is typical to use a struct in the CGNS APIs. 
Otherwise, I have to pack the parameters into an array (with type conversion) for cg_configure and then unpack the array with type conversion.

src/cgnslib.h
```diff
@@ -806,6 +839,7 @@ CGNSDLL int cg_configure(int what, void *value);
 CGNSDLL int cg_error_handler(void (*)(int, char *));
 CGNSDLL int cg_set_compress(int compress);
 CGNSDLL int cg_get_compress(int *compress);
+CGNSDLL int cg_set_filter(const struct CG_FILTER *filter);
```
brtnfld, line 842: I should probably introduce a new cg_set_chunk API, cg_configure is used currently for chunking.

src/cgnslib.c
brtnfld, line 836: This should also check if the filter is available and not just error out.

MicK7: A struct seems usable in this case, and I don't really like the type conversion stuff as it is error prone. Maybe a typedef is necessary to make the code more CGNS  like.

MicK7:
Has you are working with HDF5 filter it led me to think about encryption filtering.
I know that hdf5 is capable of encryption and I was wondering if it could be possible to have a simple encrypt for some specific data in the CGNS file (GeometryReference_t nodes only would be encrypter for instance). I know a CPEX would be required to implement this but I was still curious about the feasability.

brtnfld:
If the encryption filter is registered with HDF5, then the current API will work; one would call the encryption filter before the cgns write. However, the current implementation can't handle unregistered filters, and encryption would most likely need this.

src/cgnslib.h
brtnfld: added

brtnfld:
Need to add Fortran APIs for cg_set_chunk, cg_set_filter

src/cgnslib.c
```diff
@@ -3777,12 +3791,15 @@ int cg_section_general_write(int file_number, int B, int Z, const char * Section
     if (cgi_new_node(section->id, "ElementRange", "IndexRange_t",
         &dummy_id, data_type, 1, &dim_vals, prange)) return CG_ERROR;
 
+    HDF5storage_type = CG_CONTIGUOUS;
```
MicK7, line 3794: Nice catch

src/ptests/benchmark_hdf5.c
```diff
@@ -131,9 +139,18 @@ double t0, t1, t2;
  */
 double xtiming[15], timing[15], timingMin[15], timingMax[15];
 
+#define CD_VALUES_MAX 10
+size_t cd_nelmts=CD_VALUES_MAX;
+unsigned int cd_values[CD_VALUES_MAX];
+//hdf5_compress[0] = '\0';
+
+#ifdef H5Z_ZFP_USE_PLUGIN
```
MicK7, line 147: where is supposed to be defined H5Z_ZFP_PLUGIN
Does it come from H5pubconf.h ?

brtnfld: No, ZFP is an external filter.  It would be defined in the compiler flags, -DH5Z_ZFP_USE_PLUGIN, when the ZFP filter is available. Currently, I don't test this option in the CI tests, manually only.

MicK7:
It seems there are some conflict with the main branch.
Is there any interest in merging this branch ?

brtnfld:
I was hoping for some feedback before merging. Let me write a blurb about it on the forum and see if it gets any traction. I'll resolve the conflicts. Are you done reviewing it?

src/adfh/ADFH.c
```diff
@@ -79,6 +79,16 @@ static hsize_t h5pset_buffer_size_size            = ADFH_CONFIG_DEFAULT;
 static hsize_t h5pset_sieve_buf_size_size         = ADFH_CONFIG_DEFAULT;
 static unsigned h5pset_elink_file_cache_size_size = ADFH_CONFIG_DEFAULT;
 
+/* HDF5 Chunked Dataset Parameters */
+int chunk_ndim = 0;
+hsize_t chunk_dim[ADF_MAX_DIMENSIONS];
+
+/* HDF5 Filter Parameters */
+H5Z_filter_t filter_id;
+size_t filter_nparams;
```
gsjaardema, line 88: Should `filter_nparams` be initialized to 0, `filter_params` to `NULL` and `filter_flags` to 0?

src/adfh/ADFH.c
```diff
@@ -2913,6 +2964,8 @@ void ADFH_Put_Dimension_Information(const double   id,
   int i, swap = 0;
   hsize_t new_dims[ADF_MAX_DIMENSIONS];
   char new_type[3];
+  hid_t g_propdataset_bk;
+  int filter_set;
```
gsjaardema, line 2968: Initialize to zero?   It is done down at line 3059, but hard to see what is happening between here and there.  Should be no disadvantage to initializing here.

src/adfh/ADFH.h
```diff
@@ -162,25 +162,32 @@
...
+#define ADFH_CONFIG_ALIGNMENT              6
+#define ADFH_CONFIG_MD_BLOCK_SIZE          7
+#define ADFH_CONFIG_HDF5_BUFFER            8
 #define ADFH_CONFIG_HDF5_SIEVE_BUF_SIZE    9
 #define ADFH_CONFIG_ELINK_FILE_CACHE_SIZE 10
+#define ADFH_CONFIG_HDF5_CHUNK            11
+#define ADFH_CONFIG_HDF5_FILTER           12
 
 #define ADFH_CONFIG_RESET 800
 #define ADFH_CONFIG_RESET_HDF5 1
```
gsjaardema, line 187: This and the other setting of `ADFH_CONFIG_COMPRESS` are existing code, so probably don't conflict, but both are set to the same value of `1`.

src/tests/benchmark_hdf5.c
```diff
@@ -71,8 +71,8 @@ int  debug;
  * timing(6) = Time to read connectivity table,          20,   21,    22
  * timing(7) = Time to read solution data (field data)   23,   24,    25
  * timing(8) = Time to read array data                   26,   27,    28
- * timing(9) = Time for cgp_open, CG_MODE_WRITE          29,   30,    31
- * timing(10) = Time for cgp_open, CG_MODE_READ          32,   33,    34
+ * timing(9) = Time for cg_open, CG_MODE_WRITE          29,   30,    31
```
gsjaardema, line 74: Looks like all these are just making the comments/output correct (`cg` vs `cgp`) and not changing from testing parallel to serial.  At first it looked like changing the test, but then I saw just correcting what was printed.

gsjaardema (commented):
Looks good.  Didn't look too closely at the Fortran code, but tried to give a good look at the C.  I look forward to trying this compression.  Hopefully there will be some guidance (or pointing to the THG guidance) on setting chunk sizes particularly in parallel, but that is separate from this...

src/adfh/ADFH.c
```diff
@@ -1498,6 +1511,44 @@ void ADFH_Configure(const int option, const void *value, int *err)
...
+        chunk_ndim = 0;
+      } else {
+        val = (const cgsize_t*)value;
+        chunk_ndim = (int)(val[0]);
+      }
+      if( chunk_ndim > CGIO_MAX_DIMENSIONS ) {
+        set_error(ADFH_ERR_INVALID_USER_DATA, err);
+      }
+      for (i = 0; i < chunk_ndim; i++) {
+        chunk_dim[i] = (hsize_t)(val[i+1]);
```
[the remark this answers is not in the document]
brtnfld: chunk_ndim should be zero if val is NULL, so the loop is not executed:
lines 1519-1524.

src/adfh/ADFH.c
```diff
@@ -1498,6 +1511,44 @@ void ADFH_Configure(const int option, const void *value, int *err)
...
+        if( chunk_dim[i-1] < 1 ) {
```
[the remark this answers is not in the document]
brtnfld: fixed,thanks.

src/adfh/ADFH.c
brtnfld: good point, fixed.

src/adfh/ADFH.c
```diff
@@ -3001,15 +3056,57 @@ void ADFH_Put_Dimension_Information(const double   id,
...
+          }
+        }
+        H5Pclose(fapl); /* close the property list */
+#endif
+        /* copy the property because HDF5 does not cancel out filters when
+           using H5Z_FILTER_NONE */
+        g_propdataset_bk = H5Pcopy(mta_root->g_propdataset);
+
+        if( H5Pset_filter(mta_root->g_propdataset, filter_id,
+#if 0
```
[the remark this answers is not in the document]
brtnfld: I placed it as a future placeholder, but I'll remove it. I can't think of a scenario off-hand for when it would be mandatory.

src/adfh/ADFH.c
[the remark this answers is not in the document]
gsjaardema: Yes, agree.

MicK7:
I updated the topic to synchronize with the last develop branch. Not sure that everything went well.

PR #292, opened 2021-07-31
Cmake fixes for tools

brtnfld:
CMake fixes for CGNS-256, CGNS-257, CGNS-258


src/cgnstools/cgnsplot/CMakeLists.txt
```diff
@@ -46,7 +46,7 @@ if (WIN32)
   endif (HTML_HELP_INCLUDE_PATH AND HTML_HELP_LIBRARY)
 else (WIN32)
   include_directories(${X11_X11_INCLUDE_PATH})
-  link_libraries(${X11_X11_LIB} ${X11_Xmu_LIB} m)
+  link_libraries(${X11_X11_LIB} ${X11_Xmu_LIB} m uuid)
```
brtnfld, line 49: Not sure if I should check for uuid and add a dependency.

brtnfld:
Where are the programs vertex_to_cells and cells_to_vertex, They are in the Makefile.in but not in Cmake.

src/cgnstools/cgnsplot/CMakeLists.txt
MicK7: Instead of adding uuid, adding ${X11_SM_LIB} should be better. uuid is supposed to be set by X11 in this variable

```suggestion
  link_libraries(${X11_X11_LIB} ${X11_Xmu_LIB} ${X11_SM_LIB} m)
```

MicK7:
> Where are the programs vertex_to_cells and cells_to_vertex, They are in the Makefile.in but not in Cmake.

It seems that they should be generated from convert_location.c at compile time by defining or not VERTEX_TO_CELL. This is misleading and maybe it does not work.

src/cgnstools/cgnsplot/CMakeLists.txt
brtnfld: Thanks, fixed.

brtnfld:
I see now, one has to use make everything to build it. Do you know the reason to not build them by default? It does not get built with CMake.

I would think cgns_info should be built by default.

MicK7:
> I would think cgns_info should be built by default.

I don't know why they are omitted from cmake. Maybe it was the lack of a maintainer...

MicK7 (approved):

PR #303, opened 2021-11-02
cgnstypes: avoid defining `stat`

mathstuf:
This conflicts with `struct stat` that comes in `sys/stat.h` header with
MSVC. Older compilers may have other things, but it is just not valid to
do it without more sophisticated checks.

### DIFF
--- a/src/cgns_io.c
+++ b/src/cgns_io.c
@@ -73,10 +73,12 @@ cgns_io_ctx_t ctx_cgio = { .hdf5_access = "NATIVE",
 
 #if CG_HAVE_STAT64_STRUCT
 #ifdef _WIN32
-#define stat _stat64
+#define cgns_stat _stat64
 #else
-#define stat stat64
+#define cgns_stat stat64
 #endif
+#else
+#define cgns_stat stat
 #endif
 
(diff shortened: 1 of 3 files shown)


MicK7:
LGTM

MicK7:
@mennodeij any comment on this PR ?

mathstuf:
The solution then is to do something like `struct cngs_stat` and define *that* instead of the common name (which just bleeds over into non-cgns code).

Does Windows really not have a solution analogous to `_FILE_OFFSET_BITS=64`?

MicK7:
I tried the proposed solution in #304

mennodeij:
I introduced this `#define` to be able to open files over 2GB on windows. 
I agree that this can be handled better, as long as a test is added and passed that checks this, both in 32 bit builds, 64-bit builds, and with 32-bit and 64-bit `cgsize_t`.
It would be undesirable if tools like ParaView and Tecplot all of a sudden can’t open large files anymore on windows.

MicK7:
@mathstuf Can you bring in this PR the proposed changes that I made in the cgns_stat branch ?
It will be easier for the merge.

PR #304, opened 2021-11-05
cgnstypes: avoid defining stat

MicK7:
As suggested by @mathstuf a cgns_stat is defined

### DIFF
--- a/src/cgns_io.c
+++ b/src/cgns_io.c
@@ -73,10 +73,12 @@ cgns_io_ctx_t ctx_cgio = { .hdf5_access = "NATIVE",
 
 #if CG_HAVE_STAT64_STRUCT
 #ifdef _WIN32
-#define stat _stat64
+#define cgns_stat _stat64
 #else
-#define stat stat64
+#define cgns_stat stat64
 #endif
+#else
+#define cgns_stat stat
 #endif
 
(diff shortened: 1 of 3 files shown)


mathstuf (approved):
Thanks, this looks like it'll work a lot better.

mennodeij:
This puts the burden on the library builder to define `CG_HAVE_STAT64_STRUCT`. I miss it being defined in the CMake or configure scripts. I fear that this will re-introduce problems with large files on windows.

mathstuf:
`src/CMakeLists.txt` does the detection (as does `configure`). So that seems fine to me?

The thing is that the status quo is causing problems on Windows (namely build failures) because `#define stat …` ends up conflicting with Windows headers now doing `struct stat` themselves. This ends up with the system header "redefining" `_stat32i64` because of the ham-handed `#define stat` line. It is always better to make *new* names than to try and masquerade under names that may appear if the headers are ever updated to be conforming to any API spec.

mennodeij:
Ok, my bad, I was only looking at the changed files in this PR commit assuming CG_HAVE_STAT64_STRUCT was a newly introduced variable, but it already existed. 
And I agree with your assessment of ham-handedness, the current approach is better.

PR #307, opened 2021-11-20
Fix for CGNS-263

brtnfld:
https://cgnsorg.atlassian.net/browse/CGNS-263

This line

add_library(cgns_f2c OBJECT cgns_f.F90)

was added to address, CGNS-10. @TonyGarratt can you verify this does not break windows builds?

For reference, this was the commit:

https://github.com/CGNS/CGNS/commit/51a3fe82d3fd81302592193b8f45607f3b4dc74e


brtnfld:
Looks to break builds on macOS.

MicK7:
@brtnfld I changed the fix with a CMAKE Property. Thus the compiler fPIC option will be used when generating the object.

brtnfld:
@MicK7 looks good, thanks. 

I think all we need is to verify there is not an issue on Windows. We are not currently testing CGNS with Fortran on Windows.

MicK7 (approved):

MicK7:
@TonyGarratt Do you plan to check this change ?

brtnfld:
I'll see if I can get my windows box working to test this.

MicK7:
Maybe we could rely on intel OneAPI new compiler for testing. https://github.com/oneapi-src/oneapi-ci
It looks possible to add it to github CI : https://github.com/oneapi-src/oneapi-ci/blob/master/.github/workflows/build_all.yml

brtnfld:
Looks like it since they have fortran.

MicK7:
I don't think anyone is testing CGNS with fortran on windows recently.

MicK7:
Since the reported issue is on Linux and not windows, I think we should merge this change and create a separated issue for windows (if there is a similar problem).

PR #310, opened 2021-11-29
Copy `m_dimvals` to `dimvals`.

pvcStillinGradSchool:
As the comment states, `m_dimvals[:]` should be `0`. However, the argument passed to `H5Screate_simple()` is `dimvals`, which is an uninitialised local array.


MicK7:
@smguzik @brtnfld Do you agree with the proposed modification ?

brtnfld (approved):
LGTM

PR #311, opened 2021-12-01
Set the default file_type in cg_open only when in write mode

GrahamAshcroft:
Currently, the default file_type is set prior to calling **cgio_open_file** from **cg_open** if cgns_filetype has not been specified. This leads to the error message "cgio_open_file:not a HDF5 file - required for parallel" when an ADF based file is opened and the CGNS library has been compiled with support for HDF5 and parallel output.

By now only setting the default file_type when a new file is to be created within cgio_open_file, we can avoid the above problem without changing the default behavior of the code.

Moving the setting of the file_type into cgio_open_file breaks some of the tests that explicitly call this routine. So now, just limit the default initialization of file_type within cg_open to when a new file is to be created.

ref. CGNS-266

### DIFF
--- a/src/cgnslib.c
+++ b/src/cgnslib.c
@@ -361,6 +361,7 @@ int cg_open(const char *filename, int mode, int *file_number)
             }
             break;
         case CG_MODE_WRITE:
+            /* unlink is now done in cgio_open_file */
             /* set default file type if not done */
             if (cgns_filetype == CG_FILE_NONE)
                 cg_set_file_type(CG_FILE_NONE);
--- a/src/cgnslib.c
+++ b/src/cgnslib.c
@@ -368,10 +368,6 @@ int cg_open(const char *filename, int mode, int *file_number)
             return CG_ERROR;
     }
 
-    /* set default file type if not done */
-    if (cgns_filetype == CG_FILE_NONE)
-        cg_set_file_type(CG_FILE_NONE);
-
     /* Open CGNS file */
     if (cgio_open_file(filename, mode, cgns_filetype, &cgio)) {
         cg_io_error("cgio_open_file");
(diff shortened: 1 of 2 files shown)


MicK7:
The tricky thing is that if a user try a parallel read on an ADF file, no error will be generated since cg_open does not know if we are doing serial or parallel access.

GrahamAshcroft:
> The tricky thing is that if a user try a parallel read on an ADF file, no error will be generated since cg_open does not know if we are doing serial or parallel access.

For a parallel read I'm assuming that the file has to be opened with cgp_open? Is this correct? If this is the case then, since cgp_open explicitly sets the "expected" file type to be HDF5, the error message "cgio_open_file:not a HDF5 file - required for parallel" will occur. Does this make sense?

MicK7:
You are right, you are supposed to use cgp_open which will call cg_open and set HDF5 as default.
Your patch seems right for reading since we want to continue supporting reading ADF files.

Concerning modification of ADF files, it should be similar to the write mode. When  using a parallel compiled version of CGNS, we do not want to modify or generate ADF file. I think this is more a philosophical point of view than a down to earth reason, At some point, we want user to stop relying on ADF and switch to HDF5.

src/cgnslib.c
```diff
@@ -361,17 +361,15 @@ int cg_open(const char *filename, int mode, int *file_number)
             }
             break;
         case CG_MODE_WRITE:
-            /* unlink is now done in cgio_open_file */
+            /* set default file type if not done */
```
MicK7, line 364: Maybe keep the original comment concerning unlink

GrahamAshcroft: Fine by me.

MicK7 (approved):

MicK7:
LGTM

brtnfld (approved):

PR #313, opened 2021-12-07
CGNS-263

brtnfld:
Testing CI with Intel OneAPI on windows


brtnfld:
So close, but yet so far :grimacing:

MicK7:
@brtnfld can you synchronize this topic with the develop branch ?

brtnfld:
done.

MicK7:
Very close ... in shared mode the fortran mangling does not seems to be exported to the library while in static mode some dll may be missing when launching the test ...  still very far.

If we can remove fortran mangling and rely on iso-c binding it would be so much smoother.

MicK7:
@brtnfld can you synchronize this topic with the develop branch again ?

brtnfld:
From what I can tell CGNS_ENABLE_PARALLEL does not do anything to enable parallel, only setting CGNS_ENABLE_HDF5 and HDF5_NEED_MPI enables parallel interfaces.

brtnfld:
This could be handy 

https://github.com/mxschmitt/action-tmate

MicK7:
Can we rebase this topic after merging the "update_cmake_requirement" branch ?

MicK7:
It would be nice to rebase on current develop branch.
With a recent cmake + ninja + intel oneAPI 2024.2, it should be possible to get a clean compilation with testing.

MicK7:
CI with Ninja generator seems to work.
Combining with MSVC seems a bit more tricky

brtnfld:
@MicK7 , The  cg_array* Fortran routines are not correct since the array passed does not need to be 1D, so unfortunately, if you want to have interfaces for them like this, you would need to do so for each type to have an interface up to twelve-dimensional arrays. This is why we need an API that passes a pointer.

brtnfld:
CMake is not using cgns_fortrandll.def.in file.

MicK7:
> CMake is not using cgns_fortrandll.def.in file. Am I missing something? How are the DLLs getting set? Is cmake generating them?

I thought it was handle through the annotations: !DEC$ ATTRIBUTES DLLEXPORT::

The cgns_fortran.def is indeed missing.

MicK7:
> @MicK7 , The cg_array* Fortran routines are not correct since the array passed does not need to be 1D, so unfortunately, if you want to have interfaces for them like this, you would need to do so for each type to have an interface up to twelve-dimensional arrays. This is why we need an API that passes a pointer.
> 
> For HDF5, we generated all those interfaces (For every INTEGER, REAL KINDS) at build time to maintain older use cases, but it was not feasible to do that for F2003, where array ranks can be 128, instead of 7. So, for the F2003 APIs, we switched to passing a pointer.
> 
> The changes may also break users' code for the DataDimension argument, as before, it accepted a scalar.

Do you think it would be possible to generate for the 12 dimensions cases all interfaces like hdf5 ?

brtnfld:
> I thought it was handle through the annotations: !DEC$ ATTRIBUTES DLLEXPORT::
> 
> The cgns_fortran.def is indeed missing.

It can be done either way, but the point of having the .def file is that you don't have to repeat the same thing for every API, and that information is only needed for Windows and nothing else. We need to decide the direction to take. We did do it for every function in HDF5 but then moved to extracting them to the .def file. It is a cleaner way of doing it, IMHO.

brtnfld:
> Do you think it would be possible to generate for the 12 dimensions cases all interfaces like hdf5 ?

It should be feasible; that is the only way to overload the function to provide a way to pass a pointer without introducing a new API.

brtnfld:
Should we postpone this work until after the release? I still have some work to finish for the documentation before the release.

I plan to do another release in March or April with subfiling and compression so that we can include it then.

MicK7:
> > Do you think it would be possible to generate for the 12 dimensions cases all interfaces like hdf5 ?
> 
> It should be feasible; that is the only way to overload the function to provide a way to pass a pointer without introducing a new API. The HDF5 program that generates all the interfaces is https://github.com/HDFGroup/hdf5/blob/develop/fortran/src/H5_buildiface.F90, if you're interested.
> 
> For those generated APIs, we declare the DLL in the source code—it's easier than adding it to the .def file.

Hi, I reverted the changes for cg_array_read/write and updated the CGNSDLL stuff to be coherent with existing code in cgns_f.F90 . I agree that we should use "the cgns_fortran.def" to do it cleanly. Generating the interface for the cg_array_... API seems like a complex job that cannot be achieved for the next release and should be postponed. The full Intel OneAPI CI can also be postponed to the March-April release.

brtnfld:
Closing this in favor of #935

PR #314, opened 2021-12-09
Use intel headers instead of msvc

MicK7:


MicK7:
This should help with testing intel oneapi

brtnfld (approved):
LGTM, thanks :thumbsup:

PR #331, opened 2022-02-12
Very basic detection of duplicate face indices in NFaces

MicK7:
Basic test suggested by Beta CAE


MicK7:
The current ngon example of cgns website should fail with the new test

brtnfld (approved):
lgtm

PR #332, opened 2022-03-02
Fixed clang -Wstrict-prototypes warnings

seanm:
Clang is considering making this warning an error by default: https://discourse.llvm.org/t/rfc-enabling-wstrict-prototypes-by-default-in-c/60521


seanm:
@brtnfld BTW, you might want to consider enabling this warning on your CI, if you have a clang that supports it...

PR #333, opened 2022-03-08
CGNS-106

brtnfld:
Use macro for OpenGL-related package management.


src/cgnstools/CMakeLists.txt
```diff
@@ -22,17 +24,41 @@ if (NOT APPLE)
...
+      add_definitions(-DHAVE_OPENGL_GLU_H)
+    endif()
   else ()
     mark_as_advanced(CLEAR OPENGL_glu_LIBRARY)
     list(APPEND Missing "OPENGL_glu")
   endif ()
+
+  check_include_file(GL/glx.h HAVE_GL_GLX_H)
+  check_include_file(OpenGL/glx.h HAVE_OPENGL_GLX_H)
+  if(HAVE_GL_GLU_H)
```
MicK7, line 55: Did you mean HAVE_GL_GLX_H ?

src/cgnstools/CMakeLists.txt
```diff
@@ -2,6 +2,8 @@
 # cgnstools #
 #############
 
+add_definitions(-DCMAKE_BUILD)
```
MicK7, line 5: add_definition is not advisable in modern cmake. We should consider using target_compile_definition instead

src/cgnstools/CMakeLists.txt
```diff
@@ -22,17 +24,41 @@ if (NOT APPLE)
   include(FindOpenGL)
   if (OPENGL_FOUND)
     mark_as_advanced(FORCE OPENGL_INCLUDE_PATH OPENGL_gl_LIBRARY)
+    check_include_file(GL/gl.h HAVE_GL_GL_H)
+    check_include_file(OpenGL/gl.h HAVE_OPENGL_GL_H)
+    if(HAVE_GL_GL_H)
+      add_definitions(-DHAVE_GL_GL_H)
```
MicK7, line 30: add_definition is not advisable in modern cmake. We should consider using target_compile_definition instead

MicK7 (commented):
I don't really unsterstand the need behind the proposed change.
Indeed with cmake, the lines in tkogl/CMakeLists.txt
target_link_libraries(tkogl PUBLIC ${TKOGL_LIBS} OpenGL::GL OpenGL::GLU)
and
target_include_directories(tkogl PUBLIC ${TCL_INCLUDE_PATH} ${TK_INCLUDE_PATH} ${OPENGL_INCLUDE_DIR}) 
should be enough

Unless we want to compile without failure even when OpenGL is not available. Is it the goal ?

brtnfld:
> I don't really unsterstand the need behind the proposed change. Indeed with cmake, the lines in tkogl/CMakeLists.txt target_link_libraries(tkogl PUBLIC ${TKOGL_LIBS} OpenGL::GL OpenGL::GLU) and target_include_directories(tkogl PUBLIC ${TCL_INCLUDE_PATH} ${TK_INCLUDE_PATH} ${OPENGL_INCLUDE_DIR}) should be enough
> 
> Unless we want to compile without failure even when OpenGL is not available. Is it the goal ?

The changes are to improve the Autotools builds, I don't find anything wrong with the CMake builds. I'm trying to minimize what I need to change in CMake due to the Autotool changes.

brtnfld:
I guess at some point we should have a metal version for apple.

MicK7:
> I guess at some point we should have a metal version for apple.

I guess the major issue is that cgnsplot is still relying on old OpenGL fixed pipeline and it is required to move to a more modern dynamic pipeline with shaders in order to use OpenGL/ES,  metal or vulkan. This would be a nice and hard change.

PR #334, opened 2022-03-09
remove the use of relative paths, link_directories(.), CGNS-163

brtnfld:
removed the last use of link_directories(.)


brtnfld:
Is there a reason to keep the commented out  link_directories(.) ?

MicK7 (approved):

MicK7:
> in

I don't think so.

PR #335, opened 2022-03-15
More efficient check for duplicate faces

MicK7:
The new algorithm uses a hash to speed up the check.
It allows to not spend too much time when the file is correct.


brtnfld (approved):
lgtm

PR #338, opened 2022-03-18
increment version for next release

brtnfld:


MicK7:
I think we also need to update the VersionList at line 262 of cgnslib.c

MicK7 (approved):

brtnfld:
fixed

brtnfld:
I missed that in the 4.3.0 release, I should add 4300 and I guess release it again.

ArchangeGabriel:
> I missed that in the 4.3.0 release, I should add 4300 and I guess release it again.

Please avoid re-releasing, it breaks things in distros that had already packaged the new version (Arch Linux for instance AFAIC). Now we have two different tarballs for 4.3.0, our archiving system does not like that… Better release a 4.3.1 that just fix that and move on. ;)

ArchangeGabriel:
Also I recommend creating a release checklist for things like this that should be done at each release, so that you don’t miss any. :)

brtnfld:
You're right, my apologies.

PR #342, opened 2022-04-27
update hdf5 tag

MicK7:


brtnfld (commented):
Maybe we should define an hdf5_vers variable instead?

MicK7:
> Should the single-branch be hdf5-, with a dash?

I don't know. I think it could better to separate the tag/branch with hdf5- from the directory name hdf5_ .
But it like colors, it is just a matter of preferences.

MicK7:
The point I am not really happy with is that even when "hdf5: without" is set we need to define "hdf5_version"

MicK7:
By the way why were the tags changed on hdf5 github repository ?

brtnfld:
I thought you could set a global variable by setting it at a higher level? For example, in "on", i.e. outside of the matrix.

MicK7:
> Changed in what sense?

Before tags were hdf5_1_10_7 and were recently changed to hdf5-1_10_7 .

MicK7:
> 
> I thought you could set a global variable by setting it at a higher level? For example, in "on", i.e. outside of the matrix.
> 
> maybe it is too much trouble, and your original method is ok.
> 
> It looks like you would set
> 
> env: hdf5_vers: "x"
> 
> then you can use it by $hdf5_vers

ok done.

MicK7:
> The branches were hdf5_, and the tags were hdf5-, we generally remove branches after they are tagged. So, it is better that we change to use hdf5- instead.

ok thanks for the explanation. I understand better why the CI broke down.

brtnfld (approved):

PR #358, opened 2022-09-11
add a dummy structure for init

MicK7:


MicK7:
@brtnfld here is a dirty change. I hope it can fix nvcc issue.

brtnfld (approved):
Fixes the issue with the nvidia compiler, thanks!

MicK7:
> Fixes the issue with the nvidia compiler, thanks!
Good to know. I did not realize that it was a gcc feature.

Issue #687, opened 2023-01-17
Wrong number of bases reported for v4.3


brtnfld:
A potential issue with version 4.3 of CGNS. Attached is a simple code that opens the attached file and prints out the CGNS and HDF5 version etc.

The code is able to load the file with CGNS3.3.1:

V3 libraries:
API Version: 3.31
File is of type: HDF
HDF5 Version: 1.10.5
File Version: 3.00
Number of bases: 1
Opened and closed successfully

but when compiled against CGNS 4.3.0 and a more recent HDF5 version the result is

V4 libraries:
API Version: 4.30
File is of type: HDF
HDF5 Version: 1.12.2
File Version: 3.20
Number of bases: 0
Opened and closed successfully

i.e. it believes there are no bases. However, when the same file is opened that is based on ADF then the result is

API Version: 4.30
File is of type: ADF
HDF5 Version: 1.12.2
File Version: 3.00
Number of bases: 1
Opened and closed successfully

[issue.tar.gz](https://github.com/CGNS/CGNS/files/10437076/issue.tar.gz)

brtnfld:
Indeed, I see the same issue, and I don’t see in the tests that we actually test the value returned for the number of bases.

I think there might be a bug in cg_open. For some reason, ADFH_Number_of_Children is not calling the callback count_children.

brtnfld:
H5Literate2 is failing with:

HDF5-DIAG: Error detected in HDF5 (1.15.0) thread 0:
  #000: ../../src/H5L.c line 1698 in H5Literate2(): synchronous link iteration failed
    major: Links
    minor: Iteration failed
  #001: ../../src/H5L.c line 1663 in H5L__iterate_api_common(): link iteration failed
    major: Links
    minor: Iteration failed
  #002: ../../src/H5VLcallback.c line 5517 in H5VL_link_specific(): unable to execute link specific callback
    major: Virtual Object Layer
    minor: Can't operate on object
  #003: ../../src/H5VLcallback.c line 5483 in H5VL__link_specific(): unable to execute link specific callback
    major: Virtual Object Layer
    minor: Can't operate on object
  #004: ../../src/H5VLnative_link.c line 368 in H5VL__native_link_specific(): error iterating over links
    major: Links
    minor: Iteration failed
  #005: ../../src/H5Lint.c line 2247 in H5L_iterate(): link iteration failed
    major: Links
    minor: Iteration failed
[...]

Issue #569, opened 2023-01-02
[CGNS-173] Add backward compatibility file format test for CGNS


brtnfld:
> This issue has been imported from JIRA. Read the [original ticket here](https://cgnsorg.atlassian.net/browse/CGNS-173).

- _**Created at:**_ Wed, 18 Sep 2019 16:36:57 -0500

<p>Currently, there is no test that reads an older already existing CGNS file with a newer version of CGNS in order to verify that the data is read correctly. This test needs to be added, and a parallel test would also be useful to have.</p>

<p>Also need to test the effects of HDF5 LIB_VERSION on backward compatibility. For example, can an older CGNS library read a file that was created with LATEST libverbound of a new CGNS version if no new features were used?</p>

brtnfld:
> This comment has been imported from JIRA. Read the [original comment here](https://cgnsorg.atlassian.net/browse/CGNS-173#comment-13892).

- _**Created at:**_ Fri, 21 Feb 2020 10:37:05 -0600

<p>These tests should be added to: tests/<a href="https://github.com/CGNS/CGNS/pull/164/files#diff-ad2e33c3617ea720e20f4d6459803ade" class="external-link" rel="nofollow noreferrer">cgioc_hdf.c</a></p>

brtnfld:
Related to issue #687

PR #690, opened 2023-02-06
Fixed backward compatibilty.

brtnfld:
Fixed backward compatibility issue with H5Literate, and added backward compatibility test, with different CGNS versioned files, #687  #569


brtnfld:
Waiting on reporter to test with his file.

src/adfh/ADFH.c
```diff
@@ -2335,50 +2334,50 @@
...
-    void *tconv; void *bkg;
-    H5Pset_buffer(g_propfileopen, h5pset_buffer_size_size, tconv, bkg);
-  }
+    /* https://docs.hdfgroup.org/hdf5/develop/group___d_x_p_l.html#title16
+     * 'Sets type conversion and background buffers.
+     * 1 MByte is default.
+     */
+    if ( h5pset_buffer_size_size != ADFH_CONFIG_DEFAULT ) {
+      void *tconv; void *bkg;
+      H5Pset_buffer(g_propfileopen, h5pset_buffer_size_size, tconv, bkg);
```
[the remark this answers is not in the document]
brtnfld: fixed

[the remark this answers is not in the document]
brtnfld: fixed

MicK7:
Overall looks good to me.

MicK7 (approved):

MicK7:
I am just wondering if keeping a compile flag to force backward compatibility break with H5T_ORDER_CRT could be useful. For instance if it decided at a later time to drop backward compatibility for a performance purpose.

brtnfld:
I see. Yes, there are a lot of places still in the current implementation where it is hard-coded to enable tracking creation, and so if a CGNS file with no tracking is modified, it would probably result in tracking being enabled. So we would need to add a flag to override the current hard-coded behavior.

src/adfh/ADFH.c
```diff
@@ -2436,13 +2435,36 @@ void ADFH_Database_Open(const char   *name,
...
+    /*
+      NOTE: Creation  order was set by  default  in CGNS 3.1.3, so a
+      CGNS file created by earlier versions will not have  this set.
+      Therefore, it should not be automatically assumed to be set in
+      H5Literate.
+    */
+
     gid = H5Gopen2(fid, "/", H5P_DEFAULT);
+
+    /* Obtain the group creation flags and check for link creation ordering. */
```
MicK7, line 2448: `if (mode == ADFH_MODE_RDO) {`

brtnfld: The problem is that it makes backward compatibility dependent on the mode the file is opened. CGNS should be able to read old files, regardless of how it is opened. It should be possible to open the file for modification and not write anything.

brtnfld: We need to document that CGNS maintains one-way forward compatibility insofar as any file open and modified by, for example, 4.3.0 will be readable with 4.3.0+.  It can't be guaranteed the reverse 4.3.0- will be true.

brtnfld:
User reported this PR is able to read the CGNS file.

PR #706, opened 2023-05-05
ADFH: check for the H5L info structure type directly

mathstuf:
Newer HDF5 builds can still choose to have old ABIs.

### DIFF
--- a/src/adfh/ADFH.c
+++ b/src/adfh/ADFH.c
@@ -61,6 +61,18 @@ static int CompressData = -1;
 
 #define ADFH_CONFIG_DEFAULT 0
 
+#if H5_VERSION_GE(1,10,3) && !defined(H5_USE_18_API) && !defined(H5_USE_16_API)
+#define ADFH_HDF5_HAVE_110_API 1
+#else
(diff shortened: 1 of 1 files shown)


brtnfld (changes requested):
Thanks for noticing this issue.

All the #if H5_VERSION_GE(1,12,0) need to be replaced in the file with:

#if H5_VERSION_GE(1,12,0) && !defined(H5_USE_110_API) && !defined(H5_USE_18_API) && !defined(H5_USE_16_API)

instead of #if defined(H5Lget_info_vers) && H5Lget_info_vers == 2

It works then, for example, when hdf5 1.14 is built with v18 API compatibility.

mathstuf:
Is making a top-level define for this better at the top of the file? `CGNS_HDF5_HAVE_112_API`?

brtnfld:
> Is making a top-level define for this better at the top of the file? `CGNS_HDF5_HAVE_112_API`?

That would be fine as well.

brtnfld (approved):
LGTM, thanks!

PR #712, opened 2023-05-25
Fixed backward compatibility issue with HDF5

brtnfld:
Made it so CGNS can open and modify a CGNS file that is > HDF5 1.8 file format. It will maintain lower-bounds compatibility with the original file.


jhendersonHDF (approved):
LGTM

MicK7 (approved):

MicK7:
Thanks for adding the note in the code.
LGTM

PR #714, opened 2023-05-31
removed enabling multidataset feature, feature is still a work in progress

brtnfld:
This feature is being reworked under the multidataset branch.


MicK7:
If I understand correctly, the multidataset feature will thus not be included in 4.4 since hdf5 1.14 has API changes.
So we are again facing the problem of how a CGNS feature map to a corresponding hdf5 version.
I am starting to think that a reliable way to ask the CGNS library which features are enabled or not at runtime depending on which hdf5 version was used during compilation, which "CGNS profile" was requested (backward compatibility with CGNS xxx or not) , ... would be really a good idea.

MicK7 (approved):

PR #718, opened 2023-06-21
CMake Utilities Option and Bug Fixes

carlos-pereyra:
Added a new CMake option which enables the build of command line utilities (e.g. ``cgns_to_plot3d``, ``cgns_to_aflr3``, etc.) without having to build the entire cgnstools suite. This avoids Tk/Tcl issues, especially for folks on Mac OS X and solely want to use the command line utilities without compiling the old-school fortran conversion scripts.

Example build case would look like:
``cd CGNS/``
``mkdir build``
``cmake .. -DCGNS_ENABLE_UTILS=ON``
...
``make -j4``

Other Bug Fixes:
- Typo fix in ``cgns_to_plot3d.c``
- Unable to find HDF5_VERSION, the true variable is HDF5_VERSION_STRING.


MicK7:
Thanks for the proposal.
Could you separate each fix in a different PR please ?

carlos-pereyra:
Sure I can do that. Should we close this PR then?

MicK7:
Yes we can close it

PR #721, opened 2023-06-22
Fix bugs

carlos-pereyra:
Bug Fixes:
Typo fix in ``cgns_to_plot3d.c``
There was also a missing binaryio library ref issue within the makefile configs.

### DIFF
--- a/src/cgnstools/tkogl/tkogl.h
+++ b/src/cgnstools/tkogl/tkogl.h
@@ -22,6 +22,8 @@
 #	        define EXPORT(a,b) a b
 #	    endif
 #   endif
+#elif defined(__APPLE__)
+#include <X11/Xatom.h>
 #else
 #   define EXPORT(a,b) a b
 #   include <X11/Xatom.h>		/* for XA_RGB_DEFAULT_MAP atom */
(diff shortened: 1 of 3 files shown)


MicK7:
LGTM

MicK7 (approved):

brtnfld:
Currently, we are disabling building tools on Mac in actions. Does this fix allow enabling building tools on Mac? Though, I think cgnsview also has issues.

src/cgnstools/tkogl/tkogl.h
```diff
@@ -22,6 +22,8 @@
 #	        define EXPORT(a,b) a b
 #	    endif
 #   endif
+#elif defined(__APPLE__)
+#include <X11/Xatom.h>
```
brtnfld, line 26: Does Apple not need the EXPORT definition?

carlos-pereyra: You're correct this does not fix the compilation of ``cgnsview`` on Intel Mac OS X (Big Sur) and I'm sure if the EXPORT def would be needed to properly compile the GUI. Just a heads up, on mac I couldn't find the Xmu/stdCmap.h header so I just added that specific preprocessor directive.

I initially wanted to try compiling the viewer, but realized there were many issues and this may have been one of them. Sadly I didn't succeed getting viewer to work, so feel free to commit a revision to remove these lines.

I hope in the future y'all allow the viewer option to work on Mac OS X (or at least exist so someone else finds a fix). Many of the Tk methods used are deprecated. It would be possible to create a new version of tkogl specifically for CMake to compile for Mac OS X further down the line.

brtnfld: We are planning on fixing cgnsview on Mac for CGNS 4.5.0

PR #722, opened 2023-06-22
CMake Option for Compiling Commandline Utilities 

carlos-pereyra:
Added a new CMake option which enables the build of command line utilities (e.g. cgns_to_plot3d, cgns_to_aflr3, etc.) without having to build the entire cgnstools suite. This avoids Tk/Tcl issues, especially for folks on Mac OS X and solely want to use the command line utilities without compiling the old-school fortran conversion scripts.

Example build case would look like:
``cd CGNS/``
``mkdir build``
``cmake .. -DCGNS_ENABLE_UTILS=ON``
...
``make -j4``


CMakeLists.txt
```diff
@@ -236,7 +236,7 @@ endif ()
 # HDF5 #
 ########
 
-option(CGNS_ENABLE_HDF5 "Enable or disable HDF5 interface" "ON")
```
MicK7, line 239: HDF5 is the default backend for CGNS so this line should not be changed

CMakeLists.txt
```diff
@@ -41,7 +41,7 @@ endmacro ()
 SET_CGNS_BUILD_TYPE()
 
 project("cgns" C)
-
+include(GNUInstallDirs)
```
MicK7, line 44: If possible add a check on UNIX:

```suggestion
if(UNIX)
  include(GNUInstallDirs)
else()
  if (WIN32)
    set(${CMAKE_INSTALL_LIBDIR} "lib")
    set(${CMAKE_INSTALL_DATADIR} "share")
    set(${CMAKE_INSTALL_INCLUDEDIR} "include")
    set(${CMAKE_INSTALL_BINDIR} "bin")
    message(STATUS "Setting installation destination on Windows to: ${CMAKE_INSTALL_PREFIX}")
  else()
    message(FATAL_ERROR "System not UNIX nor WIN32 - not implemented yet")
  endif()
endif()
```

MicK7 (changes requested):

CMakeLists.txt
```diff
@@ -338,8 +350,8 @@ if (CGNS_ENABLE_HDF5)
   endif ()
 
   # Check if HDF5 version is 1.8 or greater
-  if (HDF5_VERSION VERSION_LESS "1.8.0")
```
MicK7, line 341: I haven't found any cmake documentation about HDF5_VERSION_STRING ... while HD5_VERSION should be properly defined.

brtnfld: Correct, It should be HDF5_VERSION, which is set by find_package HDF5

brtnfld:
What happens for the case CGNS_ENABLE_CGNSTOOLS=ON, CGNS_ENABLE_UTILS=OFF. Does it still build the command line tools?

CMakeLists.txt
carlos-pereyra: After updating to HDF5 version 1.10.9 the error I got doesn't occur, so yes the correct variable should be HDF5_VERSION. I'll push up the change.

carlos-pereyra: > What happens for the case CGNS_ENABLE_CGNSTOOLS=ON, CGNS_ENABLE_UTILS=OFF. Does it still build the command line tools?

The additional CMake commands should not impact the CGNS_ENABLE_CGNSTOOLS option. Yes CGNSTOOLS should be able to compile the tools. UTILS just allows for compiling the command line tools without the Tk/Tcl GUI.

brtnfld:
Can you add to the Mac github action the option -DCGNS_ENABLE_UTILS=ON?

MicK7:
Closing for now. Maybe a similar pull can be reopen in the future if the tools stay with the library.

Issue #671, opened 2023-01-02
[CGNS-276] Add missing Fortran cgp APIs


brtnfld:
> This issue has been imported from JIRA. Read the [original ticket here](https://cgnsorg.atlassian.net/browse/CGNS-276).

- _**Created at:**_ Wed, 11 May 2022 10:48:13 -0500

<ul>
	<li>cgp_coord_general_write_data_f</li>
	<li>cgp_coord_general_read_data_f</li>
	<li>cgp_poly_section_write_f</li>
	<li>cgp_poly_elements_write_data_f</li>
	<li>cgp_field_general_write_data_f</li>
	<li>cgp_field_general_read_data_f</li>
	<li>cgp_array_general_write_data_f</li>
	<li>cgp_array_general_read_data_f</li>
</ul>

<p>Include Fortran tests for the new API.</p>

brtnfld:
cgp_poly_elements_read_data_f is missing from list.

brtnfld:
Implemented and tested part of the list in PR #807.

PR #730, opened 2023-08-11
Add cgp_ptset_write/read functions

jrwrigh:
Adding parallel writing functionality for `cg_boco_write`.

Closes #728 

## Todos
- [x] Add for completion `cgp_ptset_read_data`
- [x] Add documentation
- [x] Add Fortran wrappers
- [x] Add Fortran tests of wrappers

### DIFF
--- a/src/cg_ftoc.c
+++ b/src/cg_ftoc.c
@@ -4478,17 +4478,17 @@ CGNSDLL void FMNAME(cgp_array_read_data_f, CGP_ARRAY_READ_DATA_F) (
 /*-----------------------------------------------------------------------*/
 
 CGNSDLL void cgp_ptlist_write_data_f(
-	cgint_f *fn, cgsize_t *start, cgsize_t *end, cgsize_t *points, cgint_f *ier)
+	cgint_f *fn, cgsize_t *rmin, cgsize_t *rmax, cgsize_t *points, cgint_f *ier)
 {
-    *ier = (cgint_f)cgp_ptlist_write_data((int)*fn, *start, *end, points);
+    *ier = (cgint_f)cgp_ptlist_write_data((int)*fn, *rmin, *rmax, points);
 }
 
 /*-----------------------------------------------------------------------*/
 
 CGNSDLL void cgp_ptlist_read_data_f(
-	cgint_f *fn, cgsize_t *start, cgsize_t *end, cgsize_t *points, cgint_f *ier)
+	cgint_f *fn, cgsize_t *rmin, cgsize_t *rmax, cgsize_t *points, cgint_f *ier)
 {
-    *ier = (cgint_f)cgp_ptlist_read_data((int)*fn, *start, *end, points);
+    *ier = (cgint_f)cgp_ptlist_read_data((int)*fn, *rmin, *rmax, points);
 }
 
 /*-----------------------------------------------------------------------*/
--- a/src/cg_ftoc.c
+++ b/src/cg_ftoc.c
@@ -4477,6 +4477,22 @@ CGNSDLL void FMNAME(cgp_array_read_data_f, CGP_ARRAY_READ_DATA_F) (
 
 /*-----------------------------------------------------------------------*/
 
+CGNSDLL void cgp_ptlist_write_data_f(
+	cgint_f *fn, cgsize_t *start, cgsize_t *end, cgsize_t *points, cgint_f *ier)
+{
+    *ier = (cgint_f)cgp_ptlist_write_data((int)*fn, *start, *end, points);
+}
+
+/*-----------------------------------------------------------------------*/
+
+CGNSDLL void cgp_ptlist_read_data_f(
+	cgint_f *fn, cgsize_t *start, cgsize_t *end, cgsize_t *points, cgint_f *ier)
+{
+    *ier = (cgint_f)cgp_ptlist_read_data((int)*fn, *start, *end, points);
+}
+
+/*-----------------------------------------------------------------------*/
+
 CGNSDLL void cgp_error_exit_f()
 {
     cgp_error_exit();
--- a/src/cg_ftoc.c
+++ b/src/cg_ftoc.c
@@ -3212,6 +3212,7 @@ CGNSDLL void FMNAME(cg_gopath_f, CG_GOPATH_F) (cgint_f *fn,
     CGNS_FREE(c_path);
 }
 
+
 /* - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - *\
  *              Read Multiple path nodes                                 *
 \* - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - */
@@ -4489,6 +4490,15 @@ CGNSDLL void cgp_ptlist_read_data_f(
 	cgint_f *fn, cgsize_t *rmin, cgsize_t *rmax, cgsize_t *points, cgint_f *ier)
 {
     *ier = (cgint_f)cgp_ptlist_read_data((int)*fn, *rmin, *rmax, points);
+}
+
+
+/*-----------------------------------------------------------------------*/
+CGNSDLL void  cgp_parent_data_write_f(cgint_f *fn,
+	             cgint_f *B, cgint_f *Z, cgint_f *S,
+                     cgsize_t *rmin, cgsize_t *rmax, cgsize_t *parent_data, cgint_f *ier)
+{
+    *ier = (cgint_f)cgp_parent_data_write((int)*fn, (int)*B, (int)*Z, (int)*S, (int) *rmin, (int)*rmax, parent_data);
 }
 
 /*-----------------------------------------------------------------------*/
--- a/src/cgnslib.c
+++ b/src/cgnslib.c
@@ -8868,10 +8868,12 @@ int cg_boco_write(int file_number, int B, int Z, const char * boconame,
 
      /* Save Point-Set on Disk */
     if (npnts > 0) {
+        HDF5storage_type = CG_CONTIGUOUS;
         char_33 PointSetName;
         strcpy(PointSetName, PointSetTypeName[boco->ptset->type]);
         if (cgi_write_ptset(boco->id, PointSetName, boco->ptset, index_dim,
             (void *)pnts)) return CG_ERROR;
+        HDF5storage_type = CG_COMPACT;
     }
     if (boco->location != CGNS_ENUMV(Vertex)) {
         double dummy_id;
--- a/src/cgnslib.c
+++ b/src/cgnslib.c
@@ -8869,10 +8869,8 @@ int cg_boco_write(int file_number, int B, int Z, const char * boconame,
 
      /* Save Point-Set on Disk */
     if (npnts > 0) {
-        HDF5storage_type = CG_CONTIGUOUS;
         if (cgi_write_ptset(boco->id, boco->ptset->name, boco->ptset, index_dim,
             (void *)pnts)) return CG_ERROR;
-        HDF5storage_type = CG_COMPACT;
     }
     if (boco->location != CGNS_ENUMV(Vertex)) {
         double dummy_id;
--- a/src/cgnslib.c
+++ b/src/cgnslib.c
@@ -8828,6 +8828,7 @@ int cg_boco_write(int file_number, int B, int Z, const char * boconame,
     boco->location = location;
     boco->ptset = CGNS_NEW(cgns_ptset,1);
     boco->ptset->type = ptype;
+    strcpy(boco->ptset->name, PointSetTypeName[boco->ptset->type]);
     strcpy(boco->ptset->data_type,CG_SIZE_DATATYPE);
     boco->ptset->npts = npnts;
 
@@ -8869,9 +8870,7 @@ int cg_boco_write(int file_number, int B, int Z, const char * boconame,
      /* Save Point-Set on Disk */
     if (npnts > 0) {
         HDF5storage_type = CG_CONTIGUOUS;
-        char_33 PointSetName;
-        strcpy(PointSetName, PointSetTypeName[boco->ptset->type]);
-        if (cgi_write_ptset(boco->id, PointSetName, boco->ptset, index_dim,
+        if (cgi_write_ptset(boco->id, boco->ptset->name, boco->ptset, index_dim,
             (void *)pnts)) return CG_ERROR;
         HDF5storage_type = CG_COMPACT;
     }
--- a/src/pcgnslib.c
+++ b/src/pcgnslib.c
@@ -2165,3 +2165,72 @@ int cgp_array_multi_read_data(int fn, int *A, const cgsize_t *rmin,
 
   return CG_ERROR;
 }
+
+/*===== Boundary Condition Functions =============================*/
+
+int cgp_boco_write(int file_number, int B, int Z, const char * boconame,
+          CGNS_ENUMT(BCType_t) bocotype,
+          CGNS_ENUMT(PointSetType_t) ptset_type,
+          cgsize_t npnts, int *BC)
+{
+    cgns_zone *zone;
+    cgns_zboco *zboco;
+    cgns_boco *boco = NULL;
+    int index, i, index_dim;
+    CGNS_ENUMT(PointSetType_t) ptype;
+    CGNS_ENUMT(GridLocation_t) location;
+    cgsize_t length;
+
+     /* get memory address of file */
+    cg = cgi_get_file(file_number);
+    if (check_parallel(cg)) return CG_ERROR;
+
+    if (cgi_check_mode(cg->filename, cg->mode, CG_MODE_WRITE))
+        return CG_ERROR;
+
+  return cg_boco_write(file_number, B, Z, boconame, bocotype, ptset_type, npnts, NULL, BC);
+}
+
+int cgp_boco_write_data(int file_number, int B, int Z, int bcID, cgsize_t start,
+    cgsize_t end, const cgsize_t *points)
+{
+  int elemsize;
+  hid_t hid;
+  cgns_zone *zone;
+  cgns_zboco *zboco;
+  cgns_boco *boco = NULL;
+  cgsize_t rmin, rmax;
+  CGNS_ENUMT(DataType_t) type;
+
+    /* get memory address of file */
+  cg = cgi_get_file(file_number);
+  if (check_parallel(cg)) return CG_ERROR;
+
+  if (cgi_check_mode(cg->filename, cg->mode, CG_MODE_WRITE))
+      return CG_ERROR;
+
+  boco = cgi_get_boco(cg, B, Z, bcID);
+  if (boco == 0) return CG_ERROR;
+
+  if (points) {
+    if (start > end ||
+        start < 1 ||
+        end > boco->ptset->npts) {
+      cgi_error("Error in requested element data range.");
+      return CG_ERROR;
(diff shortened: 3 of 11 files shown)


jrwrigh:
@brtnfld I have this first pass at the `cgp_boco_write*` functions. I have `cgp_boco_write` successfully creating the `ZoneBC_t`, `BC_t`, and `IndexArray_t` nodes. But when I go to `cgp_boco_write_data`, I'm getting:
```
H5Dwrite() failed
```
I've looked in gdb at the data being fed into `H5Dwrite()`, but don't notice anything completely out of the ordinary. The actual return value from the function is just `-1`, which is apparently just the default value that HDF5 returns for errors.

I'm currently handing `readwrite_data_parallel` the ID from `boco->ptset->id`, as I believe that's the ID that we want to write to. 

2 questions:
- Is `boco->ptset->id` the node we'd want to pass to `readwrite_data_parallel`?
- Do you know what could be causing the `H5Dwrite` failure?

jrwrigh:
Found one issue. The `PointSet` was a compact dataset rather than a contiguous. If I understand correctly, parallel read/writes aren't compatible with compact. I've changed all `boco_write` calls (including serial calls) to use a contiguous layout. I think this makes sense as the data stored for `PointList` items is quite large and I don't think we'd be too concerned about performance for `PointRange`. I guess we could change it based on the `PointSetType_t`?

Unfortunately, this still results in `H5Dwrite` failures.

brtnfld:
I'll take a look. You will also need to add a test for the API.

jrwrigh:
Thanks! I actually figured out the other reason it doesn't write: It doesn't like dimensions of `32 x 1`. If I force the `PointSet` to be `32` instead, it writes fine.

Specifically, I changed 
https://github.com/CGNS/CGNS/blob/d81283228f3951ace3f96827f12c487c9da96db2/src/cgns_internals.c#L8303-L8306
to
```c
    dim_vals[0]=ndim;
    dim_vals[0]=ptset->npts;
    num_dim = 1;
```

Now, whether that's actually a full fix we'd want to implement, I'm not sure. I'm not 100% sure why a `PointList` would ever need to be more than, well, a list (ie. beyond a vector), but I'm certainly not an expert on the standard.

jrwrigh:
Here's an example CGNS output from my testing for `PointList` being created as a rank 1 array rather than a rank 2 array.
[chefOut.cgns.gz](https://github.com/CGNS/CGNS/files/12324663/chefOut.cgns.gz)

brtnfld:
I'm making a branch cgp_boco to continue the work, so can you make a PR that branch when it is created?

- [ ] Add for completion cgp_boco_read
- [ ] Add tests for cgp_boco_write/read to ptest, pcgns_ctest.c
- [ ] Add documentation
- [ ] Add Fortran wrappers
- [ ] Add Fortran tests of wrappers

jrwrigh:
> Add tests for cgp_boco_write/read to ptest, pcgns_ctest.c

I was thinking about tacking this onto `ptests/test_unstruc_quad.c`. Any qualms with that?

brtnfld:
The test will need to test structured, unstructured, and mixed grid.

brtnfld:
I don't think we need cgp_boco_write since cg_boco_write accepts NULL for the point list case.

Also, your point 1 is correct, rmin and rmax are not necessarily scalars.

jrwrigh:
> I don't think we need cgp_boco_write since cg_boco_write accepts NULL for the point list case.

I think that works if all the points for a given boco are on the same rank, but not if they're spread out between multiple ranks, yes?

jrwrigh:
So `cgp_boco_write` is not necessary, but `cgp_coord_write` and `cgp_field_write` effectively does the same thing (checks the file state for parallel, then runs the normal/serial `cg_coord_write`/`cg_field_write` with a NULL input).

brtnfld:
> So `cgp_boco_write` is not necessary, but `cgp_coord_write` and `cgp_field_write` effectively does the same thing (checks the file state for parallel, then runs the normal/serial `cg_coord_write`/`cg_field_write` with a NULL input).

The issue is cg_boco_write can be used in parallel for range selection (and point selection if careful). So having a cgp_boco_write might be confusing.

I'm not sure why it was decided to create cgp_*write APIs instead of just passing NULL to delay writing the data.

jrwrigh:
> The issue is cg_boco_write can be used in parallel for range selection (and point selection if careful). So having a cgp_boco_write might be confusing.

Do we want to perhaps clarify that by naming the function(s) `cgp_boco_list_write`? Or something else to designate that this should be used for `PointList` definitions of bocos, but not `PointRange`?

jrwrigh:
I've pushed up fixes to the parallel writing functions. It now writes out 2 dimensional arrays.

I've also added a parallel boco writing to `test_unstruc_quad.c` as a test, which it writes out successfully.

brtnfld:
Something more general, cgp_ptlist_write_data, that way it can be used in other similar APIs (for example, cg_sol_ptset_write)  in the future.

jrwrigh:
Ok. If that's the case, how do we want to communicate which `PointSet` node to write to? We can't just pass in `cgns_ptset` struct for a public API (though that would work great if you meant `cgp_ptlist_write_data` to be a private function).

brtnfld:
I'm not sure what you mean. The API will write to the corresponding index number returned from the cg_write API.

jrwrigh:
For a proposed `cgp_ptlist_write_data`, all it really needs is the `cgns_ptset` struct.

Currently, the only way to get to that struct is by looking at the members of another struct (ie. `cgns_boco` or `cgns_sol`). If `cgp_ptlist_write_data` is only given an index number how should it know to call `cgi_get_boco` or `cgi_get_sol`?

There's also the fact that if we want to make it work for any `PointSet` node, the indirection to `cgns_ptset` from a parent struct can be different. Sometimes it's `struct->ptset`, sometimes it's `struct.ptset`, and sometimes, it's `struct.ptset[i]`. We'd need to handle all of those cases.

brtnfld:
Good point, we would need an extra parameter to indicate the node label, ZoneBC_t.

jrwrigh:
Ok. Other breaking situation is for things like `cgns_conn` and `cgns_1to1`, how to differentiate between whether to fill in in `ptset` or `dptset`.

Also for `cgns_hole`, how to specify which `ptset` to work with. (if I read correctly that could have multiple `ptset`s in it's struct).

jrwrigh:
I guess our only option for generalization would be to require `cg_goto`? I don't like that (in this situation or the function in general), but I guess that's what we're left with in the current API.

jrwrigh:
I've pushed up a version of `cgp_ptlist_write_data` that relies on `cg_goto` to point to the specific PointSet node to write to. It's also implemented in `test_unstruc_quad.c`.

Assuming you're good with this, I'll add to `cgi_next_node` to have all `PointSet` entries be navigable by `cg_goto`. 

Given that this is now just writing to a general `PointList` node, I don't think it needs to be tested with different `Zone_t` types since `PointList` is agnostic of any `Zone_t` type.

src/pcgnslib.c
```diff
@@ -2165,3 +2165,71 @@ int cgp_array_multi_read_data(int fn, int *A, const cgsize_t *rmin,
 
   return CG_ERROR;
 }
+
+/*===== Boundary Condition Functions =============================*/
+
+int cgp_boco_write(int file_number, int B, int Z, const char * boconame,
```
brtnfld, line 2171: Why do you still need a cgp version? It would be better not to have a  cgp version for all the other APIs that write a point list.

src/cgnslib.c
```diff
@@ -8868,10 +8869,10 @@ int cg_boco_write(int file_number, int B, int Z, const char * boconame,
 
      /* Save Point-Set on Disk */
     if (npnts > 0) {
-        char_33 PointSetName;
-        strcpy(PointSetName, PointSetTypeName[boco->ptset->type]);
-        if (cgi_write_ptset(boco->id, PointSetName, boco->ptset, index_dim,
+        HDF5storage_type = CG_CONTIGUOUS;
```
brtnfld, line 8872: Changing to contiguous could hurt performance for cases not doing partial I/O. Maybe just change it if the point list is NULL?

src/pcgnslib.c
jrwrigh: Ah, sorry. Yeah, that's left over. I wasn't intending to keep it. Fixed now.

src/cgnslib.c
jrwrigh: I was thinking either that or if `ptset->type == PointList`? I feel like having it change based on `NULL` might be more confusing, where as a consistent "PointLists are contiguous, PointRanges are compact" is a bit clearer.

Reading on ["compact vs contiguous"](https://docs.hdfgroup.org/hdf5/develop/_l_b_dset_layout.html), I'm not sure what the definition of "very small data sets" is, but I can't imagine a `PointList` being any less than a few thousand entries. Comparatively, `PointRange` should at most be like 6 entries, so that's an obvious candidate for compact.

brtnfld: 64KiB raw data is the size limit for compact storage. Setting ptset->type == PointList will work.

jrwrigh:
I've now added a reader and test for that reader (expanded `test_unstruc_quad.c`).

For the other points to fix up, where is the documentation written out? I see that you're working on a Doxygen implementation (which is awesome!) in #726, but I'm guessing that won't be merged before this one?

jrwrigh:
And for the fortran bindings, it looks like several of the `cgp_*_write_data_f` functions are written, but commented out in `cgns_f.F90`: https://github.com/CGNS/CGNS/blob/d81283228f3951ace3f96827f12c487c9da96db2/src/cgns_f.F90#L3713-L3741

Should I go ahead and write up functions for `cgp_ptlist_*_data_f`?

brtnfld:
> For the other points to fix up, where is the documentation written out? I see that you're working on a Doxygen implementation (which is awesome!) in #726, but I'm guessing that won't be merged before this one?

You can go ahead and add the doxygen docs for the new API, you can look at pcgnslib.c in the doxygen branch for examples. I'll fix it if needed when I merge the branch.

brtnfld:
For Fortran, take a look at  SUBROUTINE cg_subreg_ptset_read_f, (and write), the commented out ones are because of passing to a C void *.

jrwrigh:
I've pushed up my attempt at the Fortran bindings. They build cleanly, but there's something wrong with my local MPI install that prevents the tests from building.

src/cg_ftoc.c
```diff
@@ -4477,6 +4477,22 @@ CGNSDLL void FMNAME(cgp_array_read_data_f, CGP_ARRAY_READ_DATA_F) (
...
 
+CGNSDLL void FMNAME(cgp_ptlist_write_data_f,CGP_PTLIST_WRITE_DATA_F)(
+	cgint_f *fn, cgsize_t *start, cgsize_t *end, cgsize_t *points, cgint_f *ier)
+{
+    *ier = (cgint_f)cgp_ptlist_write_data((int)*fn, *start, *end, points);
+}
+
+/*-----------------------------------------------------------------------*/
+
+CGNSDLL void FMNAME(cgp_ptlist_read_data_f,CGP_PTLIST_READ_DATA_F)(
```
brtnfld, line 4488: Don't use FMNAME macro, the name is that given in bind(C)

jrwrigh:
Given that I can't build fortran tests for some reason, can you make a test for the reader and writer?

src/ptests/test_unstruc_quad.c
```diff
@@ -185,9 +207,46 @@ int main(int argc, char* argv[]) {
...
+  // Test file reading
+  if (cgp_open("test_unstruc_quad.cgns", CG_MODE_READ, &fn)) cgp_error_exit();
+
+  for (int i = 0; i < nelem*4; i++) elements[i] = 0;
+  start_local = comm_rank * 3 + 1;
+  end_local = start_local + 2;
+  if (cg_goto(fn, B, "Zone_t", Z, "ZoneBC_t", 1, "BC_t", 1, "PointList", 0, NULL)) cgp_error_exit();
+  if (cgp_ptlist_read_data(fn, start_local, end_local, elements)) cgp_error_exit();
+  printf("%d: %d, %d, %d\n", comm_rank, (int)elements[0], (int)elements[1], (int)elements[2]);
+
```
brtnfld, line 245: It would be good to check the read values are correct instead of printing them. I'm working on changing the tests from printing to value checking.

src/pcgnslib.c
```diff
@@ -2165,3 +2165,129 @@ int cgp_array_multi_read_data(int fn, int *A, const cgsize_t *rmin,
...
+/**
+ * \ingroup PointListData
+ *
+ * \brief Write index array to PointList in parallel.
+ *
+ * \details Must use functions in @ref AccessingANode to point to a PointSet to read from
+ *
+ * \param[in] file_number \FILE_fn
+ * \param[in] start Index of first point in local array (base 1)
+ * \param[in] end Index of last point in local array (base 1)
```
brtnfld, line 2180: we are using rmin, rmax instead of start and end, also it is the range in the file, not a local array.

jrwrigh: Ok. For local array, I meant that it's not the entire range of the array in the file (though I see how my phrasing is confusing).

src/ptests/test_unstruc_quad.c
jrwrigh: Done

src/pcgnslib.c
jrwrigh: I've gone ahead and changed it to "range in the file" since that's the consistent usage from the doxygen branch.

jrwrigh:
> If possible, instead of `cg_goto`, it would be nice to use `cg_golist` or `cg_gopath` to not rely on variadic function.

Do you mean in the `ptests/test_unstruc_quad.c`? It looks like `cg_golist` and `cg_gopath` use the same `cgi_*_posit` backend functions, so the choice of goto vs gopath vs golist seems to be user choice.

MicK7:
Yes the point is that cg_goto does not really work well in fortran. So if people look into the test to know how to write bc they could end up being confused.
The real issue is that we need to do a cg_goto because the API may be lacking. Since we typically create the node just before writing data to it, it would be nice if the cg_goto was directly done by cg_boco_write .

jrwrigh:
> The real issue is that we need to do a cg_goto because the API may be lacking. Since we typically create the node just before writing data to it, it would be nice if the cg_goto was directly done by cg_boco_write .

Do any other functions implicitly set the cgns position? That might be confusing to most users, whether the hidden cgns_posit has changed.

I think a "better" API would be to have a `cgns_node` type and have `cg_goto` and friends return that. Then have functions use that instead of the file id, base id, zone id, etc. That'd also be able to be used by these more generic functions to point directly to the node a user wants to write to. That makes it more transparent as to what is being written to. `cgns_node` could just be an integer combined with a deterministic graph traversal method. Or better yet, be a pointer to the underlying struct. Granted that would also require a very significant change to the library and would probably need to be a major version change. But I can dream. Haha

jrwrigh:
I've moved the test over to using `cg_golist`. @brtnfld Any updates on the Fortran test? Need any help with that?

brtnfld:
I will add a Fortran test as part of #671. We have one Codacy issue, but after that, I'm okay with the PR without the Fortran tests. @MicK7, are you alright with the PR?

jrwrigh:
Ok, I've fixed up that issue now.

MicK7:
I am alright with the PR. I just added fortran wrapper.
@brtnfld can you check if it is alright on your side ?

MicK7 (approved):
LGTM

src/ptests/test_unstruc_quad_f90.F90
```diff
@@ -0,0 +1,362 @@
...
+! do mapping for C side
+  allocate(pt_labels(4))
+  pt_labels(1) = C_LOC(labels(1))
+  pt_labels(2) = C_LOC(labels(2))
+  pt_labels(3) = C_LOC(labels(3))
+  pt_labels(4) = C_LOC(labels(4))
+
+!
+!---- initialize MPI
+  CALL MPI_INIT(mpi_err)
```
brtnfld, line 88: You should move the MPI_INIT before any executable lines

src/ptests/test_unstruc_quad_f90.F90
```diff
@@ -0,0 +1,362 @@
...
+  IF (comm_rank .EQ. 0) THEN
+    elements(1) = 1
+    elements(2) = 2
+  ENDIF
+  
+  CALL cgp_elements_write_data_f(F, B, Z, S, emin(1), emax(1), el_ptr, ierr)
+  IF (ierr .NE. CG_OK) CALL cgp_error_exit_f
+
+  n_boco_elems = 1
+  CALL cg_boco_write_f(F, B, Z, "LeftBC", CGNS_ENUMV(FamilySpecified), CGNS_ENUMV(PointList), n_boco_elems, null_ptr, BC, ierr)
```
brtnfld, line 268: can you use c_null_ptr instead to make sure that works?

src/ptests/test_unstruc_quad_f90.F90
```diff
@@ -0,0 +1,362 @@
...
+    end_local(1) = 0
+    el_ptr => NULL()
+  ENDIF
+
+  PRINT *, comm_rank, ":", start_local(1), " ", end_local(1)
+  indices(3) = BC
+  ierr = cg_golist(F, B, depth, pt_labels, indices)
+  IF (ierr .NE. CG_OK) CALL cgp_error_exit_f
+
+  CALL cgp_ptlist_write_data_f(F, start_local, end_local, el_ptr, ierr)
```
brtnfld, line 288: Can you also try with an allocatable array as a target and pass it to the API using C_LOC()?

src/ptests/test_unstruc_quad_f90.F90
MicK7: done

src/ptests/test_unstruc_quad_f90.F90
MicK7: it does not work with c_null_ptr

src/ptests/test_unstruc_quad_f90.F90
MicK7: done

src/ptests/test_unstruc_quad_f90.F90
brtnfld: I think you can also do:
integer(cgsize), pointer :: c_ptr
call C_F_POINTER(C_NULL_PTR, c_ptr)
and pass c_ptr.

It might need to be an array, though
integer(cgsize), pointer :: c_ptr(:)
call C_F_POINTER(C_NULL_PTR, c_ptr, [0])

I need to become more familiar with null() in Fortran. I'm surprised it works, given that the argument does not match the API interface.

MicK7: As you said we could have an interface with pure iso_c_binding like what is done for *_f03 subroutines.

But I think that it should be a dedicated task separate from this thread in order to clean the whole fortran interface.

For instance the binding to function that ends with "_f" looks confusing to me and should be better with a '_c'

brtnfld: Agreed, I am happy with the PR as is. Are you done with your changes?

MicK7: Yes, I will stop with changes now as  I happy with the current status too.

PR #736, opened 2023-10-10
Add cgns-config.cmake file.

clintonstimpson:
This adds a cgns-config.cmake file to allow projects to use find_package(cgns).
It also adjust names of hdf5 libraries to link with to use names given by find_package(HDF5).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -372,19 +372,19 @@ if (CGNS_ENABLE_HDF5)
   mark_as_advanced(CLEAR HDF5_NEED_ZLIB HDF5_NEED_SZIP HDF5_NEED_MPI)
 
   #Modern Target Library import if not defined
-  if (NOT TARGET hdf5::hdf5-${CG_HDF5_LINK_TYPE})
-    add_library(hdf5::hdf5-${CG_HDF5_LINK_TYPE} INTERFACE IMPORTED)
+  if (NOT TARGET hdf5-${CG_HDF5_LINK_TYPE})
+    add_library(hdf5-${CG_HDF5_LINK_TYPE} INTERFACE IMPORTED)
     string(REPLACE "-D" "" _hdf5_definitions "${HDF5_DEFINITIONS}")
-    set_target_properties(hdf5::hdf5-${CG_HDF5_LINK_TYPE} PROPERTIES
+    set_target_properties(hdf5-${CG_HDF5_LINK_TYPE} PROPERTIES
                           INTERFACE_LINK_LIBRARIES "${HDF5_LIBRARY}"
                           INTERFACE_INCLUDE_DIRECTORIES "${HDF5_INCLUDE_DIRS}"
                           INTERFACE_COMPILE_DEFINITIONS "${_hdf5_definitions}")
     if (CG_HDF5_LINK_TYPE STREQUAL "shared")
-      set_target_properties(hdf5::hdf5-${CG_HDF5_LINK_TYPE} PROPERTIES
+      set_target_properties(hdf5-${CG_HDF5_LINK_TYPE} PROPERTIES
                             INTERFACE_COMPILE_DEFINITIONS H5_BUILT_AS_DYNAMIC_LIB)
 
     else()
-      set_target_properties(hdf5::hdf5-${CG_HDF5_LINK_TYPE} PROPERTIES
+      set_target_properties(hdf5-${CG_HDF5_LINK_TYPE} PROPERTIES
 			    INTERFACE_COMPILE_DEFINITIONS H5_BUILT_AS_STATIC_LIB)
     endif()
   endif ()
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -577,7 +577,7 @@ add_library(cgns_static STATIC ${cgns_FILES} $<$<BOOL:${CGNS_ENABLE_FORTRAN}>:$<
 add_library(CGNS::cgns-static ALIAS cgns_static)
 # Needed to work around a CMake > 3.8 bug on Windows with MSVS and Intel Fortran
 set_property(TARGET cgns_static PROPERTY LINKER_LANGUAGE C)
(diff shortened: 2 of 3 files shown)


brtnfld (approved):
Looks good to me.

MicK7:
I am not sûre about the removal of hdf5:: namespace for cmake.

MicK7:
I am not sûre about the removal of hdf5:: namespace for cmake. 
Eitherway looks Good to me. THANKS for the improvment

brtnfld:
I'm not a CMake expert, but I assume this is needed to get the cgns-config.cmake to work, @clintonstimpson, is that correct?

clintonstimpson:
> I'm not a CMake expert, but I assume this is needed to get the cgns-config.cmake to work, @clintonstimpson, is that correct?

Yes, it's needed to get the cgns-config.cmake to work.  The names of the hdf targets being linked to cgns are now being exposed to other projects.  By using the same names found from hdf5-config.cmake, projects can include both the cgns-config.cmake and hdf5-config.cmake and names will be consistent.

humhang:
> > I'm not a CMake expert, but I assume this is needed to get the cgns-config.cmake to work, @clintonstimpson, is that correct?
> 
> Yes, it's needed to get the cgns-config.cmake to work. The names of the hdf targets being linked to cgns are now being exposed to other projects. By using the same names found from hdf5-config.cmake, projects can include both the cgns-config.cmake and hdf5-config.cmake and names will be consistent.

hello, which hdf5-config.cmake are you using? The [FindHDF5.cmake](https://cmake.org/cmake/help/latest/module/FindHDF5.html) shipped with cmake does not import targets named hdf5-${CG_HDF5_LINK_TYPE} ("hdf5-static" or "hdf5-shared").

From my observation hdf5-${CG_HDF5_LINK_TYPE} is an interface target added by CMakeLists.txt of CGNS, it needs to be exported and added in cgns-config.cmake.

clintonstimpson:
> hello, which hdf5-config.cmake are you using? The [FindHDF5.cmake](https://cmake.org/cmake/help/latest/module/FindHDF5.html) shipped with cmake does not import targets named hdf5-${CG_HDF5_LINK_TYPE} ("hdf5-static" or "hdf5-shared").
> 
> From my observation hdf5-${CG_HDF5_LINK_TYPE} is an interface target added by CMakeLists.txt of CGNS, it needs to be exported and added in cgns-config.cmake.

The `hdf5-config.cmake` I'm referring to is created if you compile hdf5 from source code using CMake, and is supplied by the hdf5 developers.  I'm looking at the FindHDF5.cmake supplied by the CMake developers and see that it provides namespace prefixed target names, so they aren't exactly the same.  Hmm...

Are you proposing cgns-config.cmake export the target, or are you proposing that cgns-config.cmake resolve the target by using FindHDF or hdf5-config?

humhang:
> > hello, which hdf5-config.cmake are you using? The [FindHDF5.cmake](https://cmake.org/cmake/help/latest/module/FindHDF5.html) shipped with cmake does not import targets named hdf5-${CG_HDF5_LINK_TYPE} ("hdf5-static" or "hdf5-shared").
> > From my observation hdf5-${CG_HDF5_LINK_TYPE} is an interface target added by CMakeLists.txt of CGNS, it needs to be exported and added in cgns-config.cmake.
> 
> The `hdf5-config.cmake` I'm referring to is created if you compile hdf5 from source code using CMake, and is supplied by the hdf5 developers. I'm looking at the FindHDF5.cmake supplied by the CMake developers and see that it provides namespace prefixed target names, so they aren't exactly the same. Hmm...
> 
> Are you proposing cgns-config.cmake export the target, or are you proposing that cgns-config.cmake resolve the target by using FindHDF or hdf5-config?

I think the best way is to import HDF5 via find_dependency in the cgns-config.cmake, so that user only needs to invoke find_package(cgns).

In the find_dependency we can specify using modulus mode (through FindHDF5.cmake) or Config mode (through hdf5-config.cmake). We can try config mode first so that we do not need to export the hdf5-xxx targets. and if it fails we can fall back to modulus mode, and export that target.

clintonstimpson:
Agreed, I think find_dependency would be a nice addition to the cgns-config.cmake.
Is it possible to avoid exporting HDF5 targets - even as a fall back?
Perhaps if we also adjust the CGNS CMake file to link with the target names provided by the FindHDF5.cmake file, then we could avoid the need to export the targets ourselves.  FindHDF5 provides targets hdf5::hdf5, hdf5::hdf5_hl, etc...

clintonstimpson:
Here's some additional data...  I see that CMake 3.19 added support for targets in FindHDF5.cmake.  Since CGNS has a minimum requirement of 3.12, we might need to continue supporting a FindHDF5.cmake without targets.

humhang:
> Agreed, I think find_dependency would be a nice addition to the cgns-config.cmake. Is it possible to avoid exporting HDF5 targets - even as a fall back? Perhaps if we also adjust the CGNS CMake file to link with the target names provided by the FindHDF5.cmake file, then we could avoid the need to export the targets ourselves. FindHDF5 provides targets hdf5::hdf5, hdf5::hdf5_hl, etc...

Yes, I think this could be the best solution but this means a lot changes needed in the CMakeLists.txt of CGNS.

humhang:
> Here's some additional data... I see that CMake 3.19 added support for targets in FindHDF5.cmake. Since CGNS has a minimum requirement of 3.12, we might need to continue supporting a FindHDF5.cmake without targets.

I think this doesn’t really matter. When using HDF5 we do not explicitly link our target against targets like hdf5::* but ${HDF5_LIBRARIES} which could be targets or absolute paths to HDF5 depending on CMake version. If it is target then find_dependency(HDF5) in the cgns_config.cmake will import that anyway.

MicK7:
I'd like to add my two cents. Since CGNS cmake version requirement mainly follow HDF5 choices, we could update the requirement to cmake 3.18. Would it simplify the problem ?

PR #738, opened 2023-10-26
Fix for when using HDF5 1.14.3 and onward with NAG compiler.

brtnfld:
IEEE standard arithmetic enables software to raise exceptions such as overflow, division by zero, and other illegal operations without interrupting or halting the program flow. The HDF5 C library intentionally performs these exceptions. Therefore, the "-ieee=full" nagfor switch is necessary when compiling a program/tests with CGNS using the NAG compiler to avoid stopping on an exception.


MicK7:
Should we also add flags for gfortran (like "-fno-unsafe-math-optimizations -frounding-math -fsignaling-nans")  ?

MicK7 (approved):

PR #742, opened 2023-12-05
fix uninitialized warning

MicK7:
When updating SU2, this piece of code raised the compiler warning.
Since it is a simple fix. I upstream the patch.


brtnfld (approved):
lgtm

Issue #749, opened 2024-01-24
Building CGNS tools with Intel LLVM compilers fails


mennodeij:
We're preparing for changing from the "Classic" Intel compilers to the LLVM-based ones. These are in the oneAPI toolkit:
`CC=icx CXX=icpx FC=ifx F90=ifort`

When the build is configured with CMake and -DCGNS_BUILD_CGNSTOOLS=ON  (default), there are many build warnings like this:
`warning: a function definition without a prototype is deprecated in all versions of C and is not supported in C2x [-Wdeprecated-non-prototype]`

and build errors:
```
src/cgnstools/tkogl/tkogl.c:602:8: error: call to undeclared function 'TkWmAddToColormapWindows'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
       TkWmAddToColormapWindows(tkwin);
```

If `CGNS_BUILD_CGNSTOOLS=OFF` is chosen, the build completes successfully without any warnings or errors.

It would be nice if this new Intel LLVM compiler was supported for the CGNS tools as well.

brtnfld:
- [ ] Check if #757 addresses the issue.

brtnfld:
The error has been fixed int he current version, the warning is addressed by #910

PR #757, opened 2024-03-03
backport gcc14 fedora patch

MicK7:


MicK7:
This patch could also help with #749

MicK7:
@mennodeij could you confirm if this patch also solves your oneapi issue ?

MicK7:
> @MicK7 sorry, I can't. New job, no CGNS.

All the best on your new adventure! Thanks for taking the time to answer.

brtnfld (approved):
LGTM

src/cgnstools/tkogl/tkogl.c
```diff
@@ -22,6 +22,12 @@
 #include "printstr.h"
 #include "feedback.h"
 
+#if ! defined(__WIN32__) && ! defined(_WIN32)
+/* For TkWmAddToColormapWindows. */
+#define _TKPORT /* Typical installations cannot find tkPort.h. */
+#include <tk-private/generic/tkInt.h>
```
brtnfld, line 28: This breaks opensuse builds, where tkInt.h is in /usr/include

brtnfld: We'll have to see how Kitware is handling it:

https://github.com/Kitware/VTK/blob/dbd25545a57c976b15a95f4651df17bc84540f41/Rendering/Tk/vtkTkRenderWidget.cxx#L21

I did a little searching for Autotools, but definitely no great solutions for finding it other than using pkg-config or requiring the TCL/TK_INCLUDE_DIR variable, as I don't think providing TKdir is enough to find it. Or cycle through a list of common directories where it is found.

PR #759, opened 2024-03-08
updated cg_goto_f to use F2008 API

brtnfld:
Experimenting with switching to using F2008 features.


brtnfld:
@MicK7, I wonder why we are passing an array of strings as an argument for cg_goto_f. I don't think passing an array would work (at least, I don't think any tests do that).

I'm unsure why we did not "flip the switch" to doing it this way; it was probably a compiler issue.

It also looks like we never technically use the end variable, so if someone does "end=end," it won't work.

brtnfld:
The goto_f interface should be DIMENSION(1) and not DIMENSION(*), as passing a character array with a len != 1 does not work (and is probably not common anyway since no one has complained).

It should be overloaded to handle a case where a vector is not passed.

```
CHARACTER(C_CHAR), DIMENSION(*), INTENT(IN), OPTIONAL :: UserDataName1,UserDataName2, &
            UserDataName3,UserDataName4,UserDataName5,UserDataName6,UserDataName7,UserDataName8, &
            UserDataName9,UserDataName10,UserDataName11,UserDataName12,UserDataName13,UserDataName14, &
            UserDataName15,UserDataName16,UserDataName17,UserDataName18,UserDataName19,UserDataName20
```

brtnfld:
@MicK7, What is the issue with the MacOS runner?

MicK7:
> @MicK7, What is the issue with the MacOS runner?

It seems the MacOS runner image swithed from Apple clang 14 to Apple clang 15.
In the build gcc is specified through the CC variable but it is well apple clang that is used for the build.
It appears that gfortran 13 may not link properly with clang 15 abi or there is something missing at compile and/or link time.

MicK7:
This branch can also be rebased

MicK7:
This branch needs to be rebased

MicK7:
It seems that the call to cg_goto in fortran 2008 was not tested correctly and was passed over to cg_ftoc.c. Mapping with the C function cg_goto_08 is a problem. The issue with array vs not array of character seems to be a consequence of C API but we need to keep a more natural API for the fortran user.

brtnfld:
Are you saying an array of characters is the natural API?

MicK7:
> Are you saying an array of characters is the natural API?

As a fortran user I would say the opposite. The natural API whould be CHAR(LEN=...) DIMENSION(1)
But ISO C seems to need CHAR(LEN=1) DIMENSION(*) that's why the goto_f08 is problematic.

MicK7:
I moved some subroutine from cg_ftoc to cgns_f.F90 to see what was the more natural mapping to keep existing interface

brtnfld:
> As a fortran user I would say the opposite. The natural API whould be CHAR(LEN=...) DIMENSION(1)
> But ISO C seems to need CHAR(LEN=1) DIMENSION(*) that's why the goto_f08 is problematic.

Right, I would be astonished if anyone in Fortran is passing an array of char(LEN=1) array. The CGNS documentation does not take a stand on the structure of the string being passed, but all the tests and examples don't use character arrays. 
It would be much better if it specified that the Fortran character arguments should not be arrays.

MicK7:
@brtnfld I will not go further with the cgns_f.F90 changes. I wanted to start the work in this PR but to do the whole file need a dedicated PR IMO and a bit of AI scripting to go faster. Now we really need to check that it works with all modern fortran compiler.

MicK7 (approved):

brtnfld:
It's a start at least. We hope to do the same with HDF5, but with over 200+ APIs, it's a sizeable, tedious task. 
I'll run the branch through our testing here before merging.

brtnfld:
@MicK7, it passes the test we run here. However, we are not testing Fortran on Windows, so we must test these changes before merging or releasing them. Have you managed to get Fortran/intel to work in CI on Windows? It has been a while since I looked into it.

MicK7 (approved):

PR #761, opened 2024-04-16
Update cmake requirement

MicK7:
- Update cmake requirement to be coherent with hdf5.
  This could also help with packaging on the long run.

- Fix Fortran to C mapping on Apple ARM64
 This is compiler dependent so a better design based on ISO C should be possible.
 The current correction is based on https://gcc.gnu.org/onlinedocs/gcc-8.1.0/gfortran.pdf  (see hidden parameter section)

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -112,6 +112,10 @@ jobs:
             "--${{ matrix.parallel }}-parallel --${{ matrix.fortran }}-fortran --${{ matrix.hdf5 }}-hdf5 --${{ matrix.tools }}-cgnstools"
       shell: bash
 
+    - name: Setup tmate session
+      uses: mxschmitt/action-tmate@v3
+      if: matrix.os == 'macos-latest'
+
 ##################################
 # TEST CGNS (CMAKE)
 ##################################
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -119,8 +119,8 @@ jobs:
     - name: CMake test CGNS
       run: |
         cd cbuild
-        make -j 8
-        make test
+        cmake --build . --config "Debug" --parallel 8
+        ctest -C "Debug" --verbose --output-on-failure
       shell: bash
 
 ##################################
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -33,7 +33,7 @@ jobs:
             artifact: "macOS.tar.xz"
             os: macos-latest
             parallel: disable
-            fortran: with
+            fortran: without
             hdf5: without
             tools: disable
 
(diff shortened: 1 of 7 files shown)


.github/workflows/main.yml
```diff
@@ -20,7 +20,7 @@ jobs:
   build:
     strategy:
       matrix:
-        name: ["Ubuntu Latest GCC", "macOS Latest Clang"]
+        name: ["Ubuntu Latest GCC", "macOS Latest"]
```
brtnfld, line 23: Maybe you should use macos-13, which is x86, so we can still test Fortran on Macs.

MicK7: Indeed a separate runner is needed for x86 Mac. But my goal is to achieve M1/M2/M3 compatibility with Fortran enabled. I have deactivated temporary Fortran and will try to solve the issue as soon as possible. So far, I suspect that transfer of strings between C and Fortran on arm64 may be part of the problem... and I have not figured out why.
I will try to enable an interactive github action to investigate if possible.

brtnfld: Ok, thanks. I did not realize you were still looking into it. I think flang is now being used, which could be part of the issue. flang was at least an issue for HDF5 Fortran on macs in general.

MicK7: @brtnfld Problem solved \o/

The solution works with clang and gcc.

MicK7:
@brtnfld Please review

The branch use clang bug gcc has also been tested (I kept commented the github action line to do it)
Flang is not tested and will require a bit more investigation.

CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.12)
+cmake_minimum_required(VERSION 3.20)
```
brtnfld, line 1: Is there a CMake feature that requires the version bump?

MicK7: Yes, it is to be coherent with recent HDF5 and also from this cmake version the linking process to HDF5 (finding dependencies, packaging to transfer to consumer code the dependency) has been modified to be more reliable (feature https://cmake.org/cmake/help/v3.20/module/FindHDF5.html).

appveyor.yml
```diff
@@ -33,9 +33,9 @@ skip_tags: true
 #---------------------------------#
 environment:
   global:
-    1_8: &1_8 1.8.21
-    1_10: &1_10 1.10.6
-    1_12: &1_12 1.12.0
+    1_8: &1_8 1.8.23
+    1_10: &1_10 1.10.10
+    1_12: &1_12 1.12.3
```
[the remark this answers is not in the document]
MicK7: Thanks for the review. I replaced 1.12 by 1.14. Now waiting for the CI to complete to validate this change.

brtnfld (approved):
LGTM, thanks!

PR #771, opened 2024-07-07
[pcgns] add poly elements read/parent elements readwrite

ndelling:
Hi,

this branch includes following improvements of the pcgns library:

- add `cgp_poly_elements_read_data_offsets` and `cgp_poly_elements_read_data_elements` to read unstructured element connectivity in parallel (note that the process is divided in two parts, first read the offsets to allocate elements array on user side, and then read elements)
- add `cgp_parentelements_read_data` and `cgp_parentelements_write_data` to read/write a ParentElements node without reading/writing the ParentElementPosition node (only ParentElements has to be read in a lot of cases, and, when applying some renumbering methods on the mesh, only ParentElements may have to be overwritten).
- one missing function in Fortran API
- tests for the new functions


src/cgns_f.F90
```diff
@@ -3724,6 +3724,22 @@ END SUBROUTINE cgp_section_write_f
 !!$       INTEGER(CGSIZE_T) , INTENT(IN) :: elements
 !!$       INTEGER, INTENT(OUT) :: ier
 !!$     END SUBROUTINE cgp_elements_write_data_f
+!!$
+     SUBROUTINE cgp_poly_section_write_f(fn, B , Z, sectionname, type, start, end, maxoffset, nbndry, S, ier) !BIND(C, name="cgp_poly_section_write_f")
```
MicK7, line 3728:
```suggestion
 SUBROUTINE cgp_poly_section_write_f(fn, B , Z, sectionname, type, start, end, maxoffset, nbndry, S, ier)
    IMPLICIT NONE
    IMPORT :: cgsize_t, cgenum_t
    INTEGER, INTENT(IN) :: fn
    INTEGER, INTENT(IN) :: B
    INTEGER, INTENT(IN) :: Z
    CHARACTER(LEN=*), INTENT(IN) :: sectionname
    INTEGER(cgenum_t), INTENT(IN) :: type
    INTEGER(cgsize_t), INTENT(IN) :: start
    INTEGER(cgsize_t), INTENT(IN) :: end
    INTEGER(cgsize_t), INTENT(IN) :: maxoffset
    INTEGER, INTENT(IN) :: nbndry
    INTEGER, INTENT(OUT) :: S
    INTEGER, INTENT(OUT) :: ier
    
    CHARACTER(LEN=LEN_TRIM(sectionname)+1,KIND=C_CHAR) :: c_name
    INTERFACE
       INTEGER FUNCTION cgp_poly_section_write(fn, B , Z, sectionname, type, start, end, maxoffset, nbndry, S) BIND(C, name="cgp_poly_section_write")
       IMPORT :: c_char, cgsize_t, cgenum_t
       IMPLICIT NONE
       INTEGER, INTENT(IN) :: fn
       INTEGER, INTENT(IN) :: B
       INTEGER, INTENT(IN) :: Z
       CHARACTER(KIND=C_CHAR), DIMENSION(*), INTENT(IN) :: sectionname
       INTEGER(cgenum_t), INTENT(IN) :: type
       INTEGER(cgsize_t), INTENT(IN) :: start
       INTEGER(cgsize_t), INTENT(IN) :: end
       INTEGER(cgsize_t), INTENT(IN) :: maxoffset
       INTEGER, INTENT(IN) :: nbndry
       INTEGER, INTENT(OUT) :: S
       END FUNCTION cgp_poly_section_write
    END INTERFACE
    
    c_name = TRIM(sectionname)//C_NULL_CHAR

    ier = cgp_poly_section_write(fn, B , Z, c_name, type, start, end, maxoffset, nbndry, S)
END SUBROUTINE cgp_poly_section_write_f
```

I am not a fan of adding a commented BIND as it is misleading for developers.
I suggest a different way of adding the subroutine that should not require any work in cg_ftoc.c
Or if we really want to keep code in cg_ftoc.c, I would prefer having a cgp_poly_section_write_f_c in cg_ftoc.c without the untested FMNAME stuff and binding it from cgns_f.F90

@brtnfld what's your POV compared to what HDF5 is doing.

ndelling: @MicK7 I usually create Fortran API the same way as you did, but I wanted to match the philosophy of the other functions. Thx for the correction.

Without cg_ftoc.c and using iso_c_binding, it might also be easier to deal with null pointers and pass C_NULL_PTR to the C API when a MPI rank has no data to write. Currently, gcc sanitizer throws errors when trying to pass a null pointer to cg_boco_write_f for example, while some compilers allow it. Something like:
```
integer(cgsize_t), pointer :: array(:)
type(c_ptr) :: c_array
c_array = C_NULL_PTR
if (associated(array)) c_array = c_loc(array)
ierr = cg*_toto(c_array)
```
could do the trick. But that's another discussion.

brtnfld: This is how HDF5 does it, though you don't need the intents in the interface since C will not honor them. As I've mentioned, the issue that complicates doing it this way is the vague CGNS documentation. The Fortran character string can be an array of, for example, single characters at each element. This would not be allowed in HDF5.

I started adding the missing Fortran cgp APIs, #671, in my fork, https://github.com/brtnfld/CGNS/tree/CGNS-276, but it is still a work in progress.

brtnfld: Can you update your code to use  @MicK7 model? You should change the integer arguments being passed to the C API to C_INTs, I would also remove the intents in the interface.

ndelling: Tell me if it's what you meant

src/cgns_f.F90
```diff
@@ -4580,26 +4581,70 @@ END SUBROUTINE cgio_set_label_f
...
+      INTEGER(cgsize_t) :: start
+      INTEGER(cgsize_t) :: end
+      INTEGER(cgsize_t) :: maxoffset
+      INTEGER(c_int) :: nbndry
+      INTEGER(c_int) :: S
+      END FUNCTION cgp_poly_section_write
+    END INTERFACE
+
+    c_name = TRIM(sectionname)//C_NULL_CHAR
+    ier = cgp_poly_section_write(fn, B, Z, c_name, type, start, end, maxoffset, nbndry, S)
```
brtnfld, line 4641: You will have to, for example, INT(fn, C_INT).

ndelling: Okay

src/cgns_f.F90
```diff
@@ -4638,7 +4644,15 @@ END FUNCTION cgp_poly_section_write
     END INTERFACE
 
     c_name = TRIM(sectionname)//C_NULL_CHAR
-    ier = cgp_poly_section_write(fn, B, Z, c_name, type, start, end, maxoffset, nbndry, S)
+    c_fn = INT(fn, c_int)
+    c_B = INT(B, c_int)
+    c_Z = INT(Z, c_int)
+    c_nbndry = INT(nbndry, c_int)
```
brtnfld, line 4650: No need to create extra variables for intent in variables. Also just do ier = INT( cpg...,)

src/cg_ftoc.c
```diff
@@ -4383,6 +4383,25 @@ CGNSDLL void FMNAME(cgp_section_write_f, CGP_SECTION_WRITE_F) (
 
 /*-----------------------------------------------------------------------*/
```
brtnfld, line 4385: You can remove this.

brtnfld:
Can you also add the new Fortran functions to cgns_fortrandll.def.in

brtnfld:
We will also need a test for the Fortran API.

src/cgns_f.F90
```diff
@@ -4580,26 +4581,74 @@ END SUBROUTINE cgio_set_label_f
...
+      INTEGER(c_int) :: fn
+      INTEGER(c_int) :: B
+      INTEGER(c_int) :: Z
+      CHARACTER(KIND=C_CHAR), DIMENSION(*) :: sectionname
+      INTEGER(cgenum_t) :: type
+      INTEGER(cgsize_t) :: start
+      INTEGER(cgsize_t) :: end
+      INTEGER(cgsize_t) :: maxoffset
+      INTEGER(c_int) :: nbndry
+      INTEGER(c_int) :: S
```
brtnfld, line 4637: This is correct, without VALUE, C expects a pointer.

MicK7:
Looks very solid to me. @brtnfld can we merge ?

PR #778, opened 2024-09-06
Remove `access()` calls

jrwrigh:
From the man page for `access()`:
> Use of these functions is discouraged since by the time the returned information is acted upon, it  is  out-of-date.  (That is, acting upon the information always leads to a time-of-check-to-time-of-use race condition.) An application should instead attempt the action itself and handle the [EACCES] error that occurs if the file is not accessible

We've run into these race conditions on a few different file systems now, where when trying to write a file in parallel from a large number of parts we get 
```
CGNS error 1 cgio_open_file:File Open Error: NEW - File already exists
```

I think this is particularly prevalent on "performance" file-systems which are often not POSIX compliant. Though technically these race conditions could happen in serial too.

I've also removed the `stat == "UNKNOWN"` case to avoid the `ACCESS` calls there as well. I could not find a single place in the code where `"UNKNOWN"` would be passed to `ADFH_Database_Open` (or `ADF_Database_Open` either).

### DIFF
--- a/src/adfh/ADFH.c
+++ b/src/adfh/ADFH.c
@@ -2180,15 +2180,7 @@ void ADFH_Database_Open(const char   *name,
   for (i = 0; buff[i]; i++)
     buff[i] = TO_UPPER(buff[i]);
 
-  if (0 == strcmp(buff, "UNKNOWN")) {
-    if (ACCESS(name, 0))
-      mode = ADFH_MODE_NEW;
-    else if (ACCESS(name, 2))
-      mode = ADFH_MODE_RDO;
-    else
-      mode = ADFH_MODE_OLD;
-  }
-  else if (0 == strcmp(buff, "NEW")) {
+  if (0 == strcmp(buff, "NEW")) {
     mode = ADFH_MODE_NEW;
   }
   else if (0 == strcmp(buff, "READ_ONLY")) {
--- a/src/adfh/ADFH.c
+++ b/src/adfh/ADFH.c
@@ -2189,24 +2189,12 @@ void ADFH_Database_Open(const char   *name,
       mode = ADFH_MODE_OLD;
   }
   else if (0 == strcmp(buff, "NEW")) {
-    if (!ACCESS(name, 0)) {
-      set_error(REQUESTED_NEW_FILE_EXISTS, err);
-      return;
-    }
     mode = ADFH_MODE_NEW;
   }
   else if (0 == strcmp(buff, "READ_ONLY")) {
-    if (ACCESS(name, 0)) {
-      set_error(REQUESTED_OLD_FILE_NOT_FOUND, err);
-      return;
-    }
     mode = ADFH_MODE_RDO;
   }
   else if (0 == strcmp(buff, "OLD")) {
-    if (ACCESS(name, 0)) {
-      set_error(REQUESTED_OLD_FILE_NOT_FOUND, err);
-      return;
-    }
     mode = ADFH_MODE_OLD;
   }
   else {
--- a/src/cgns_io.c
+++ b/src/cgns_io.c
@@ -590,53 +590,71 @@ int cgio_check_file (const char *filename, int *file_type)
     int n;
     char buf[32];
     FILE *fp;
+    double rootid;
     static char *HDF5sig = "\211HDF\r\n\032\n";
     struct cgns_stat st;
 
     int mpibuf[2], err = CGIO_ERR_NONE;
 
+    *file_type = CGIO_FILE_NONE;
+
+#if 0
+    /* ACCESS call deactivated file open try-error strategy used instead */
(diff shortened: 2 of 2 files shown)


brtnfld (dismissed):
You could still have access in serial cases, but POSIX calls (including access) can be problematic for object storage systems like DAOS. Removing this is okay because the later calls will fail if the mode is unsuitable; it might not be as obvious.

MicK7:
> You could still have access in serial cases, but POSIX calls (including access) can be problematic for object storage systems like DAOS. Removing this is okay because the later calls will fail if the mode is unsuitable; it might not be as obvious.

We are dealing with a public API behavor change here. Do we have some tests or should we add one to check failure cases due to r/w access or file absent or file not being hdf5 ? 

The later calls you are speaking about are calls to "hdf5::isAccessible" or "hdf5::isHdf5" ?

brtnfld:
Yes, those and the H5F calls should fail if the mode asked for is not possible. I've not looked at the testing, but we should have tests of these various parameters.

Because we lack tests, I wonder if this is a behavior change. I would expect it to fail later on since I think these "access" tests are more of a sanity check to avoid errors down the line. For example, if an API tries to open a read-only file, HDF5 will throw an error if the file does not exist. This will probably trigger a different error code, so in that case, it would be an API change.

jrwrigh:
> This will probably trigger a different error code, so in that case, it would be an API change.

Yeah, `H5Fcreate` will return a nonzero return code if it can't work with the file. CGNS [catches that and returns](https://github.com/CGNS/CGNS/blob/c451b7fcb3017f591a9192306749af883dfe9bd8/src/adfh/ADFH.c#L2402) `FILE_OPEN_ERROR`. It'd be nice if it returned the actual HDF5 error message though in these cases.

brtnfld:
Let me see if I can add a test and check to see if there is any behavior difference other than failing differently. But I don't necessarily think this API change needs to wait until a major release.

brtnfld:
Adding a test that tries to open a nonexistent file currently results in the program hanging, which is not ideal 👎 

By the way, I've added a CGNS configure option to enable printing the HDF5 error stack upon an error. It would be handy for debugging without recompiling the source with debugging enabled for ADFH.

brtnfld:
The option will be available in a new PR I'm working on. 

Also, the changes in this PR will not affect opening or modifying a nonexistent file because it will first encounter the 'access' check in  cgio_check_file before the check in ADFH; I wonder why we need two checks.

jedbrown:
`access` is always racy if the calling code has logic depending on the result. Suppose an archival job is running in the background, moving files to slower storage after a certain age. Then your production job checks `access` and sees the file, but it's gone by the time it tries to open it. It is likely that `access` will be formally deprecated in a future version of POSIX. I think CGNS is important enough that it should not rely on such race conditions, and instead correctly handle errors from the likes of `H5Fopen`.

MicK7:
> Also, the changes in this PR will not affect opening or modifying a file because it will first encounter the 'access' check in cgio_check_file before the check in ADFH; I wonder why we need two checks.

The historical logic was call to cgio_check_file to determine if the file is accessible and the type of file by reading magic number.
Then if the filetype is ADF or NONE(ie not provided) and the type found is different from it the reading process switch to the determined type and do not crash the user application. The user is thus able to read and modify the file seamlessly.
I think that logic was implemented to help people switch from adf to hdf in their application.
Nowadays, I think the logic could be kept only for serial or one process cases. But it should completely be skipped when HDF5 is explicitly requested. Thus we would have only one check in this case by directly calling ADFH_Database_open .

jedbrown:
FWIW, that is still a race condition in serial. It might not trigger frequently because people aren't super likely to have a concurrent job (possibly on a different machine that accesses the same network file system) moving/replacing files, but that does happen. Even if the file still exists between the probe and open call, it could have been replaced by one with a different file type (ADF vs HDF5). It's kind of unfortunate there is no H5F open with file descriptor to avoid such races.

Perhaps the logic can be updated to attempt `H5Fopen` and if that fails, try with ADF (realizing the file could have been replaced or deleted between those function calls)?

MicK7 (approved):

brtnfld:
Give me some time to do some additional testing before this goes in. I should be able to get around to it by the end of the week.

MicK7:
@brtnfld  I tried to remove ACCESS call in cgio_check_file.
The strategy is now to not do any check when HDF5 is requested.
Otherwise a check is done starting by trying to open the file as hdf5 and then falling back to old fopen check with magic header.

MicK7:
I did a reset of the approval to not merge it by error before @brtnfld feedback

brtnfld:
I'm sorry, I ran out of time last week; I'll review it this week.

brtnfld (approved):
I added a test (I will commit it in a separate PR), and the error return code for a missing file for read and modify access is the same as before.

MicK7 (approved):

PR #782, opened 2024-09-23
fixed type for MPI_Allgather

brtnfld:
Since the mpi arguments are long, use MPI_LONG as the type. This fixes failures when cgsize_t is an int.


MicK7 (approved):
Indeed, it looks like it was meant to handle two cases but offset_sizes was hardcoded as long whatever happens. Let's keep it simple for now.

PR #787, opened 2024-10-30
Doxygen Section detailing CGNS File Navigation Illustration

brtnfld:


src/doxygen/CGNSNav.dox
```diff
@@ -0,0 +1,44 @@
...
+ *       
+ * \code{C}
+ *   cg_goto(filenum, basenum, "Zone_t", 1, "UserDefinedData_t", 1, NULL);
+ *   cg_goto(filenum, basenum "Zone1", 0, "UserDefinedData_t", 1, NULL);
+ *   cg_goto(filenum, basenum, "Zone_t", 1, "User", 0, NULL);
+ *   cg_goto(filenum, basenum, "Zone1", 0, "User", 0, NULL);
+ *   cg_gopath(filenum, "/Base/Zone1/User");
+ * \endcode
+ * Now, to change to the user-defined data node under zone 2, you may use the full path
+ * specification as above, or else a relative path, using one of the following:
```
MicK7, line 20: Maybe add a comment here that " The next relative cg_gorel/cg_gopath calls are valid only if cg_goto or cg_gopath with an absolute were used once before ! "

MicK7 (approved):

Issue #800, opened 2024-12-04
Address CodeQL  report.


brtnfld:
At minimum the critical issues so that the CI passes, see report here:
https://github.com/CGNS/CGNS/pull/799/checks?check_run_id=33933370399

brtnfld:
Closing as addressed by PR #953 (merged 2026-05-15), which resolved the CodeQL format specifier and overflow warnings in `cgnslib.c`. The CodeQL CI is now passing.

PR #801, opened 2024-12-04
Address CodeQL issues, #800

brtnfld:


src/cgnslib.c
```diff
@@ -7016,6 +7021,11 @@ int cg_elements_general_write(int fn, int B, int Z, int S,
 
         if (read_parent_data(section)) return CG_ERROR;
 
+        if(cnt > CG_SIZE_MAX/newsize ||
+           (cnt*newsize) > CG_SIZE_MAX/sizeof(cgsize_t) ) {
```
MicK7, line 7025:
```suggestion
        if((cnt*newsize) > CG_SIZE_MAX/sizeof(cgsize_t) ) {
```

MicK7 (dismissed):
Looks good with a small test that can be simplified

PR #807, opened 2024-12-19
Added Fortran Particle APIs, other missing Fortran APIs, and Tests

brtnfld:
Called the C APIs directly from Fortran
Reworked cg_configure_f to accept callbacks.

Misc. C clean-up

- [ ] Check Fortran DLL with new APIs.
- [x]  Update the Doxygen with Fortran specific information as needed.
- [x]  Overload the function with min-max arguments that can be an array or scalar.
- [x]  We will need to find a solution to libgl1-mesa-glx libglu1-mesa-dev being removed from ubuntu 24.
- [x]  Missing fortran APIs: `cgp_poly_elements_read_data_elements(_f)`, `cgp_poly_elements_read_data_offsets(_f)`,`cgp_parentelements_read_data(_f)`

Related to #772 final tasks.

PR #808, opened 2024-12-20
fix CI testing missing gl libs

brtnfld:


brtnfld:
I'm checking if PR #807 caused parallel tests to fail. There seems to be some issue with Ubuntu 24 that is unrelated to my changes.

brtnfld:
I think it is this issue: https://bugs.launchpad.net/ubuntu/+source/mpich/+bug/2072338

When I run the parallel program, it lists the com rank as 0 and the comm size as 1  for all the ranks.

brtnfld:
fixed in PR #807

PR #823, opened 2025-01-11
update install cmake process

MicK7:
- use same command as the documentation


brtnfld (approved):
We should add a Codacy configuration file and ignore md files.

PR #831, opened 2025-02-25
remove spurious space in Fortran C-binding

melven:
Breaks compilation with nvfortran...


MicK7 (approved):

MicK7:
@melven Thanks a lot for the patch.

PR #834, opened 2025-04-08
find_package(ZLIB) for compatibility with more recent hdf5

andrewcorrigan:
When I build CGNS with more recent versions of HDF5 I get the error message shown below.   This adds a call to `find_package(ZLIB)` that detects the `ZLIB::ZLIB` target.

```
CMake Error at /CGNS/build/CMakeFiles/CMakeTmp/cmTC_417f5Targets.cmake:21 (set_target_properties):
  The link interface of target "hdf5-static" contains:

    ZLIB::ZLIB

  but the target was not found.  Possible reasons include:

    * There is a typo in the target name.
    * A find_package call is missing for an IMPORTED target.
    * An ALIAS target is missing.

Call Stack (most recent call first):
/CGNS/build/CMakeFiles/CMakeTmp/CMakeLists.txt:15 (include)


CMake Error at src/CMakeLists.txt:388 (try_compile):
  Failed to generate test project build system.
Call Stack (most recent call first):
  src/CMakeLists.txt:476 (CHECK_HDF5_FEATURE)
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -345,7 +345,6 @@ if (CGNS_ENABLE_HDF5)
 
   set(HDF5_NEED_ZLIB "OFF" CACHE BOOL "Does the HDF5 library require linking to zlib?")
   if(HDF5_NEED_ZLIB)
-    find_package(ZLIB)   # more recent versions of HDF5 require ZLIB::ZLIB
     find_library(ZLIB_LIBRARY z)
     mark_as_advanced(CLEAR ZLIB_LIBRARY)
   else ()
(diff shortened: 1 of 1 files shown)


MicK7:
Looks like there was a recent change in hdf5 too: https://github.com/HDFGroup/hdf5/pull/5280
This part of the code is 15years old and may be needed only when hdf5 is not built with cmake.
I wonder if it is still needed.

MicK7:
Could you create an issue with the steps to reproduce the bug ?
With information:
- hdf5 version
- how hdf5 was built (from source or provided package from linux distro) ?
- cmake version
And if possible share your hdf5-config.cmake (in case hdf5 was built with cmake tooling).

MicK7:
@andrewcorrigan Thanks for the proposed changes. I modified your solution to get a more automatic detection.
You can check if it works for you.

HDF5 should provide all the correct linking information for cmake to work without the user getting involved.
I suspect that something similar could be necessary for szip.

MicK7:
@brtnfld Do you know if it can be solved at the hdf5 packaging level ?

andrewcorrigan:
> @andrewcorrigan Thanks for the proposed changes. I modified your solution to get a more automatic detection. You can check if it works for you.

I tested it using [SEACAS](https://github.com/sandialabs/seacas) with [this patch](https://github.com/sandialabs/seacas/pull/632) removed and it's working.

MicK7 (approved):

PR #842, opened 2025-04-21
Modifiy cg_configure to retrieve a value

### LINKED ISSUE #840: Proposal: Expose ADFH_MAXIMUM_FILES in client accessible include
For robustness in my application which reads multiple files output from a parallel execution in file-per-rank mode (A mpi run on 1024 ranks will create 1024 files), I need to know whether I can have all the input CGNS files open at the same time, or whether I need to access then in a round-robin manner keeping only a subset open.

I can query the underlying operating system to determine its limit on the number of open files `sysconf(_SC_OPEN_MAX)` or `_getmaxstdio()`, but that limit may exceed the internal limit set by `ADFH_MAXIMUM_FILES`.

The problem is that my application has no way to know what the value of `ADFH_MAXIMUM_FILES` is, so I cannot reliably check that limit.  The current value is 1024 which is the same or greater than the limit on many operating systems, but it would still be good to be able to verify that limit.

Can we move the definition of `ADFH_MAXIMUM_FILES` to a user-visible include file (cgnslib.h, cgnsconfig.h, ...) such that it can be queried at build time by an application.

Thanks.

brtnfld:
Updates `cg_configure` source, tests, and documentation to retrieve a value. Added new option to get  from the HDF5 interface `ADFH_MAXIMUM_FILES`

Closes #840


gsjaardema (dismissed):
Looks good.  My only minor complaint is that most uses of `cg_configure` are to set values and this is the first use to just get values.  There may be some confusion in that a user might think they can change the maximum files using this interface.  The use of `GET` in the `CG_CONFIG_GET_MAXIMUM_FILES` helps elminate that confusion though.  LGTM.

MicK7:
As @gsjaardema said the use of cg_configure looks a bit confusing. 
Why do we need an API ? Would it be a bad practice to expose this value as a global public parameter of the library ?

```C
static int cg_config_maximum_files = ADFH_MAXIMUM_FILES ;
```

brtnfld:
I was trying to avoid new API expansion and use the parameter, but if that would cause confusion, I can introduce a cg_get_configure instead. But it sounds like the preference would be to add a global parameter, although we haven't been exposing adfh information into upstream headers or vice versa.

gsjaardema:
> I was trying to avoid new API expansion and use the parameter, but if that would cause confusion, I can introduce a cg_get_configure instead. But it sounds like the preference would be to add a global parameter, although we haven't been exposing adfh information into upstream headers or vice versa.

I am good with the current solution.  It is slightly confusing, but the name of the parameter with the `_GET_` in the name does help the confusion.  I am also good with exposing the value though.

MicK7 (dismissed):

brtnfld:
Added a doxygen note reinforcing the idea of limited retrieval options for cg_configure

PR #848, opened 2025-05-12
Improve HDF5 cmake handling

MicK7:
The cmake logic is improved.
Indeed, CGNS only needs to look for ZLIB or SZIP with libaec libraries when hdf5 is static.
If HDF5 is shared those dependencies remain PRIVATE and will only be linked by HDF5.

The new cmake logic try yo find package only if they are not bundled with hdf5. It means that when HDF5 was built ZLIb was an external package. This package is known and provided by the compilation environment thus HDF5 do not care to give CGNS information and that's fine.
The added "find_package" may not be needed for application that wrap CGNS build with their own cmake stuff (like paraview)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -390,7 +390,7 @@ if (CGNS_ENABLE_HDF5)
     endif()
   endif ()
 
-  # If HDF5 needs ZLIB, but ZLIB is an external pakage not provided by hdf5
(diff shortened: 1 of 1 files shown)


MicK7:
@andrewcorrigan Can you check that this solution works for you

andrewcorrigan:
I see you already merged it, but it looks fine

PR #850, opened 2025-06-06
Increment library version for next release

brtnfld:
Major version update due to the addition of the particle feature.


MicK7:
Should we add an "experimental" way to ignore version for reading or at least to set a version for writing at the same time ?
I know that the change to v5 is due to introducing ParticleZone_t. But at the same time, when using an earlier version of the library, the nodes are just ignored. This is a different issue than the v4 switch.
I think we should let the user decide to set the writing version to v4.6 to make it easier for people to migrate.
It would be the user responsibility to ensure that only feature compatible with v4 series are being used.
And even if ParticleZone_t are written by error it will not crash.

brtnfld:
Yes, as was discussed briefly in the steering committee meeting, we should introduce an API similar to HDF5's "H5Pset_libver_bounds" where the default is to use the earliest library unless a feature requires a newer library version.

https://support.hdfgroup.org/documentation/hdf5/latest/_accessibility.html

MicK7 (approved):

PR #858, opened 2025-07-05
Fix cgnsview HDF5 loading

MicK7:
CGNSView is calling cgio_check_file multiple times... and it closes the file while loading it and building the tree.
The call is used by file_stats mainly but also other parts to get the file type (adf or hdf5). The CGNSview is not coded to be aware that it is requesting the file type of an already open file...
The CGIO level API has a similar issue. It tries to open the file to get it's type while not being aware that HDF5 library is  reusing an opened file and then just close it.
The propose fix is to do a less strong check.

### DIFF
--- a/src/adfh/ADFH.c
+++ b/src/adfh/ADFH.c
@@ -2514,9 +2514,15 @@ void ADFH_Database_Valid(const char   *name,
         *err = NULL_STRING_POINTER;
     else
 #if ADFH_HDF5_HAVE_112_API
-        *err = H5Fis_accessible(name, H5P_DEFAULT);
+	if (H5Fis_accessible(name, H5P_DEFAULT) <=0)
+	  *err = ADFH_ERR_NOT_HDF5_FILE;
+	else
+	  *err = NO_ERROR;
 #else
-        *err = H5Fis_hdf5(name);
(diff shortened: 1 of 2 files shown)


MicK7:
@charlest-uk Indeed there is a tricky bug with cgnsview. Thanks for reporting

MicK7:
@brtnfld I don't remember exactly why we needed a hard check by opening and closing file. Is it because H5Fis_hdf5 is not reliable ?

brtnfld:
> @brtnfld I don't remember exactly why we needed a hard check by opening and closing file. Is it because H5Fis_hdf5 is not reliable ?

H5Fis_hdf5 is deprecated because it only works with the native VOL connector, while is_accessible works with any VOL connector, such as DAOS. I'm not sure why cgns_io does not use those functions instead of opening and closing the file.

Plus, is_accessible has the advantage in parallel because it checks with one rank and broadcasts the result. The current approach involves all ranks opening and closing the file, which is not the best practice.

MicK7:
@brtnfld Thanks for the feedback. I modified the code to use is_accessible and it seems to fix the cgnsview issue.

MicK7:
@brtnfld Could you review this patch ?

brtnfld (approved):
LGTM 👍

PR #861, opened 2025-07-25
Remove debug message in cmake

jspanchu:
This pull request includes a minor change to the `src/CMakeLists.txt` file. The change removes an unnecessary status message related to the detection of `int64_t`. 

It was first reported in VTK - https://gitlab.kitware.com/vtk/vtk/-/issues/19745


MicK7:
The message was intended for debug purpose. We should have declared it as VERBOSE or DEBUG instead of STATUS

MicK7:
I don't see any issue in removing it for final user. @brtnfld what is your opinion ?

brtnfld (approved):
Seems good to me.

PR #869, opened 2025-10-03
Reinvestigated #851

### LINKED ISSUE #851: find_package(cgns) will not work with the current cgns-config.cmake and targets.
user codes are not able to use cgns simply by find_package(cgns) and target_link_libraries(xxx cgnslib) in their CMakeLists.txt. 

Because the cgns cmake target depends on the cmake target hdf5-static(shared), which is added by the CMakeLists.txt of cgns but not exported.

cgns target has interface dependency on the added target hdf5-static(shared), this target must also be exported and add before the cgns target in cgns-config.cmake so that user code can call target_link_libraries(xxx cgnslib) and use cgns trivially.

========
Test
1. user code cmakelists.txt

Image

2. cmake configuration

Image

3. compiling. Failed since hdf5-shared is not exported and thus find_package(cgns) does not import it. The interface dependency hdf5-shared of cgnslib will simply be interpreted as flag -lhdf5-shared.

Image

brtnfld:
The original code had nested conditions that prevented HDF5 from being found in all cases:

* For shared HDF5: Only searched if HAVE_LK_HDF5_SHARED was set
* For static HDF5: Only searched if explicitly requesting the static component

This meant that when a user did find_package(CGNS) without specifying components, the HDF5 dependency might not be resolved before cgns-targets.cmake was included, resulting in a linking error.

Creates the appropriate hdf5-static or hdf5-shared target based on how CGNS was configured and happens before include(cgns-targets.cmake), so the targets exist when needed

Fixes #851 

----

> [!IMPORTANT]
> Fixes HDF5 dependency resolution in `src/cgns-config.cmake.in` to prevent linking errors when using `find_package(CGNS)` without components.
> 
>   - **Behavior**:
>     - Ensures HDF5 targets (`hdf5-static` or `hdf5-shared`) are created before including `cgns-targets.cmake` in `src/cgns-config.cmake.in`.
>     - Fixes linking error when `find_package(CGNS)` is used without specifying components.
>   - **Dependency Resolution**:
>     - For shared HDF5, sets `HDF5_USE_STATIC_LIBRARIES OFF` and creates `hdf5-shared` target if not found.
>     - For static HDF5, sets `HDF5_USE_STATIC_LIBRARIES ON` and creates `hdf5-static` target if not found.
>     - Adds ZLIB and SZIP dependencies if required by HDF5.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=CGNS%2FCGNS&utm_source=github&utm_medium=referral)<sup> for c91562a062a3c285373f8f253bdadca4a838510e. You can [customize](https://app.ellipsis.dev/CGNS/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/src/cgns-config.cmake.in
+++ b/src/cgns-config.cmake.in
@@ -65,59 +65,45 @@ foreach (comp IN LISTS ${CGNS_PACKAGE_NAME}_FIND_COMPONENTS)
   endif ()
 endforeach ()
 
-# Propagate HDF5 dependency
-set (HAVE_LK_HDF5_SHARED 0)
-if ((${CGNS_PACKAGE_NAME}_ENABLE_HDF5_SUPPORT) AND (${CGNS_PACKAGE_NAME}_HDF5_LINK_TYPE STREQUAL "shared"))
-  set (HAVE_LK_HDF5_SHARED 1)
-endif()
-if (HAVE_LK_HDF5_SHARED)
-  find_dependency(HDF5)
-  if (HDF5_FOUND)
-    # Handle case of partial hdf5 cmake config
-    if (NOT TARGET hdf5-shared)
-      message (STATUS " HDF5 was found but not hdf5-shared target, trying a fall back")
+#-----------------------------------------------------------------------------
+# Resolve HDF5 dependency BEFORE including cgns-targets.cmake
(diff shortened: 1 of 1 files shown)


MicK7:
It looks cleaner indeed.
LGTM

MicK7 (approved):

Issue #779, opened 2024-09-11
CGNS MLL Bug when using 32-bit cgsize_t


brtnfld:
External report from Mark:

Our build of the MLL uses the 32-bit “cgsize_t” since our meshes have dimensions well below the 32-bit limit.  A colleague reported a CGNS MLL error when reading an R8 array with dimensions [12000, 401, 401].  This is below 2^31, so I was surprised when the MLL failed to read arrays under GridCoordinates_t.
 
I traced the problem to cg_array_read() and this declaration of “num”:
 
https://github.com/CGNS/CGNS/blob/c451b7fcb3017f591a9192306749af883dfe9bd8/src/cgnslib.c#L15193
 
The value of “num” is formed as the product of array dimensions (3 in my case) so it should be larger than cgsize_t (which is “int” in the 32-bit mode).  If the product of all dimensions exceeds the 32-bit limit, the code then fails to allocate as the size of the allocation is computed like this:
 
(size_t)( num*size_of(array->data_type) )
 
Since “num” is “int” and the CGNS “size_of” is “int”, the result of their product above is still an int which can clearly rollover for.
 
As a result of this defect, the CGNS maximum dimensions are limited to something significantly less than 2^31 (or 2^32), at least by a factor of 8 less for R8 types.
 
The fix seems easy.  Just redeclare “num” as “size_t” instead in affected cgnslib.c procedures.  Certainly, array_read, array_read_as, but I think I see a few other locations where the same size_t() defect is found.
 
Regards,
Mark
 
P.S. To reiterate, I was always under the impression that the 32-bit integer mode of CGNS supported zones where each dimension could be 2^31.  But this defect requires the product of all 3 dimensions divided by sizeof(array data type) to be less than 2^31 which is significantly less.

brtnfld:
Follow-up report from Mark:

Some further clarification.  I now see that cgi_check_dimensions() is used by the “32-bit” MLL build to assert that the product(dims)<2^31 not the individual dimensions (I was mistaken about this).  So the total zone node/vertex count must not exceed 2^31 when using the “32-bit” MLL.
 
Regardless, the noted coding in array_read_*() and other locations is effectively reducing that limit by a factor of 4 or 8 because it is multiplying “num” by an int value of 4 or 8.

MicK7:
[...]

PR #878, opened 2025-10-14
Switch to using size_t for memory-related calculations instead of cgs…

brtnfld:
Switch to using size_t for memory-related calculations instead of cgsize_t

Further investigation into #779

----

> [!IMPORTANT]
> Switches memory-related calculations from `cgsize_t` to `size_t` in `cgnslib.c` for better memory handling.
> 
>   - **Type Change**:
>     - Replace `cgsize_t` with `size_t` for memory calculations in `cg_array_read` and `cg_array_read_as` in `cgnslib.c`.
>     - Adjust type casting in `memcpy` and `malloc` calls to use `size_t`.
>   - **Function Updates**:
>     - Update `cgi_convert_data` call in `cg_array_read_as` to cast `num` to `cgsize_t`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=CGNS%2FCGNS&utm_source=github&utm_medium=referral)<sup> for b47f5d9596681cb3a38c716bb89bdd961713d5db. You can [customize](https://app.ellipsis.dev/CGNS/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/src/cgnslib.c
+++ b/src/cgnslib.c
@@ -17512,7 +17512,7 @@ int cg_array_read(int A, void *Data)
     array = cgi_array_address(CG_MODE_READ, 0, A, "dummy", &have_dup, &ier);
     if (array==0) return ier;
 
-    for (n=0; n<array->data_dim; n++) num *= array->dim_vals[n];
+    for (n=0; n<array->data_dim; n++) num *= (size_t)array->dim_vals[n];
 
     if (array->data)
         memcpy(Data, array->data, num*size_of(array->data_type));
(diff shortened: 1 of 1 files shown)


src/cgnslib.c
```diff
@@ -17515,7 +17515,7 @@ int cg_array_read(int A, void *Data)
     for (n=0; n<array->data_dim; n++) num *= array->dim_vals[n];
```
MicK7: computation of num should handle explicit conversion of cgsize_t to size_t here in this case

MicK7 (approved):

PR #906, opened 2025-11-20
Remove global hdf5_access reset

### LINKED ISSUE #836: Calling `cgp_close` resets the PARALLEL status
I think that the following code will cause problems if there is more than one file opened in parallel IO mode:
```C
int cgp_close(int fn)
{
    /* reset parallel access */
    strcpy(ctx_cgio.hdf5_access,"NATIVE");
    return cg_close(fn);
}
```

If I close one of the files, but am still writing/reading to/from the second file, then it looks like subsequent reads will not correctly set the correct COLLECTIVE mode so there will be different behavior on a write/read call if `cgp_close` is called between calls...

This also points to the need to have the `hdf5_access` be a file-specific property and not a library global property...

brtnfld:
Fixes #836: removes the global hdf5_access reset, which corrupts parallel file operations, and adds a test.

### DIFF
--- a/src/adfh/ADFH.c
+++ b/src/adfh/ADFH.c
@@ -2292,15 +2292,16 @@ void ADFH_Database_Open(const char   *name,
 #endif
 
   /* open the file */
-  access_mode = CGIO_NATIVE_MODE;
+  /* Convert format string to enum for internal use (Issue #836) */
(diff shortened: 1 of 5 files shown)


brtnfld:
Codacy is a false positive; it is being used.

MicK7 (approved):
LGTM

PR #913, opened 2025-12-04
Fix SonarQube buffer overflow warnings in parent data handling

brtnfld:
Address buffer reuse issues identified by SonarQube static analysis:

1. Out-of-bounds array access in parent data handling (cgnslib.c:)
   - Code allocated newelems for cnt*newsize elements, then reused buffer for parface needing 2*newsize
   - SonarQube flagged potential overflow if buffer sizes differ
   - Fixed by freeing parelem buffer before allocating parface buffer
   - Maintains same memory footprint while satisfying static analysis

2. Out-of-bounds array access in poly element parent data (cgnslib.c:7757-7789)
   - Same as above  in a different function
   - Same fix as above.

Note: cnt should always equal 2 per CGNS specification, but this approach makes
the buffer size transitions explicit and clearer to static analysis tools, and clearer for developers.

### DIFF
--- a/src/adf/ADF_internals.c
+++ b/src/adf/ADF_internals.c
@@ -2769,7 +2769,7 @@ default : /** Multiple data-chunks to free.  Free them,
    if( *error_return != NO_ERROR ) {
       free( data_chunk_table ) ;
       return ;
-      }
+   }
 
     /** Free each entry in the table **/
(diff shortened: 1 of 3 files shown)


brtnfld:
This was the last of the "blockers".

MicK7 (approved):

PR #920, opened 2025-12-27
Introduces an opaque parameter object (cg_parameters_t) API for thread-safety

brtnfld:
This introduces an opaque parameter object (cg_parameters_t) API to provide thread-safe configuration and forward compatibility, in preparation for version-bound handling for issue #915.

Key features:

    Defined opaque handle, version constants, and lifecycle API (create/destroy/copy).
    Utilized a save/restore pattern in open calls to ensure thread safety.
    Added test_parameters.c validation suite and updated CMake and autotools configuration.
    Included comprehensive Doxygen documentation and design rationale.

This commit resolves critical thread-safety issues in parallel I/O and enables progressive API enhancement through polymorphic macros.

Thread-Safety Fixes:
- Eliminate race condition in parallel I/O by removing global hdf5_access_mode state
- Implement Extended Internal API pattern: cgio_open_file_with_mode() for internal use with explicit mode parameter, preserving public cgio_open_file() signature

Polymorphic Macro Enhancement:
- Enable cg_open() and cgp_open() to accept 3 or 4 arguments using C99 variadic macros
  * 3-arg: cg_open(file, mode, &fn) -> uses CG_PARAMS_DEFAULT
  * 4-arg: cg_open(file, mode, params, &fn) -> uses provided params
- Implement argument-counting dispatch pattern
- Add MSVC preprocessor compatibility workaround (CG_OPEN_EXPAND macro)

----

> [!IMPORTANT]
> Introduces `cg_parameters_t` API for thread-safe configuration in CGNS, enhancing parallel I/O operations and adding comprehensive tests and documentation.
> 
>   - **Behavior**:
>     - Introduces `cg_parameters_t` for thread-safe configuration in `cgnslib.c` and `pcgnslib.c`.
>     - Adds `cg_params_create`, `cg_params_destroy`, and `cg_params_set` for managing parameter objects.
>     - Updates `cg_open` and `cgp_open` to support polymorphic macros for 3 or 4 arguments.
>   - **Thread-Safety**:
>     - Removes global `hdf5_access_mode` state in `cgns_io.c` to fix race conditions.
>     - Implements `cgio_open_file_with_mode` for explicit mode handling.
>   - **Testing**:
>     - Adds `test_parameters.c` and `test_params_fortran.F90` for validating parameter API.
>     - Updates `CMakeLists.txt` and `Makefile.in` to include new tests.
>   - **Documentation**:
>     - Adds Doxygen comments for new API functions in `cgnslib.h` and `pcgnslib.h`.
[...]

PR #927, opened 2026-01-23
feat: Update cg_golist args to be const char **

jrwrigh:
Input to `cg_golist` does not change the array or the strings, therefore can be const. I end up needing to do something like:
```c
      const char *labels[] = {"Zone_t", "ZoneBC_t", "BC_t", "PointList"};
      ierr = cg_golist(cgid, base, 4, (char **)labels, (int[]){zone, 1, BC, 0});
```

because the complier complains that the `labels` initializer dropping the `const` nature of the strings, and also that passing `const char **` to a `char **` in `cg_golist` also drops the `const` nature as well.

The change will allow me to go full C99 compound literal as well, which is nice:
```c
      ierr = cg_golist(cgid, base, 4, (const char *[]){"Zone_t", "ZoneBC_t", "BC_t", "PointList"}, (int[]){zone, 1, BC, 0});
```

whereas before I'd have needed a double cast:
```c
      ierr = cg_golist(cgid, base, 4, (char **)(const char *[]){"Zone_t", "ZoneBC_t", "BC_t", "PointList"}, (int[]){zone, 1, BC, 0});
```
---
Note that the addition of `const` to the function parameters will not break any user code; implicit cast of `char **` -> `const char **` will not return a warning since it's going from more to less permissive (rather than the other way around, which would return the warnings I mention above).

----

> [!IMPORTANT]
> Update `cg_golist`, `cgi_update_posit`, and `cgi_set_posit` to use `const char **` for labels, allowing C99 compound literals without extra casting.
> 
>   - **Function Signature Update**:
>     - Change `cg_golist` in `cgnslib.c` to use `const char **` for `label` parameter.
>     - Change `cgi_update_posit` and `cgi_set_posit` in `cgns_internals.c` to use `const char **` for `label` parameter.
>   - **Header Update**:
>     - Update `cgnslib.h` and `cgns_header.h` to reflect the `const char **` change in function declarations.
>   - **Backward Compatibility**:
>     - Implicit cast from `char **` to `const char **` is allowed, ensuring no breakage in existing user code.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=CGNS%2FCGNS&utm_source=github&utm_medium=referral)<sup> for 21b440c8a96617ceec0143144809a2b8586b1a05. You can [customize](https://app.ellipsis.dev/CGNS/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/src/cgns_header.h
+++ b/src/cgns_header.h
@@ -1149,8 +1149,8 @@ CGNSDLL cgns_pgoverning *cgi_get_particle_governing(cgns_file *cg, int B, int P)
 CGNSDLL cgns_pequations *cgi_get_particle_equations(cgns_file *cg, int B, int P);
 
 /* find position lead by the goto function */
-CGNSDLL int cgi_update_posit(int cnt, int *index, char **label);
-CGNSDLL int cgi_set_posit(int fn, int B, int n, int *index, char **label);
+CGNSDLL int cgi_update_posit(int cnt, int *index, const char **label);
+CGNSDLL int cgi_set_posit(int fn, int B, int n, int *index, const char **label);
 CGNSDLL int cgi_posit_id(double *posit_id);
 CGNSDLL cgns_posit *cgi_get_posit(void);
 CGNSDLL int cgi_posit_index_dim(void);
--- a/src/cgns_internals.c
+++ b/src/cgns_internals.c
@@ -13885,7 +13885,7 @@ static int cgi_next_posit(char *label, int index, char *name)
     return CG_NODE_NOT_FOUND;
 }
 
-int cgi_update_posit(int cnt, int *index, char **label)
+int cgi_update_posit(int cnt, int *index, const char **label)
 {
     int n, ierr;
     double pid, id;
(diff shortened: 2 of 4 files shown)


brtnfld:
In C, char ** and const char ** are incompatible pointer types (C11 §6.5.16.1).  GCC with standard warnings (-Wall) treats this as an ERROR with -Wincompatible-pointer-types

brtnfld:
@MicK7, do you think C11 and _Generic are widespread enough that we can adopt it? I used _Generic to address #920. but that is still a WIP.  IMHO, this is exactly what _Generic is useful for, I'm just not certain of the C11 support these days, although, HDF5 2.0 requires C11

jrwrigh:
Ach! Didn't realize the const casting gets more complicated with double pointers vs single pointers. 

I would think that there would be a type that both `char **` and `const char **` can safely cast into, but I can't seem to find one that GCC is ok with. Even `const char *const *const` won't allow safe casting into, which seems ridiculous to me (if I promise that absolutely nothing about what you're passing in will change, why should you care about the const-ness of the input?). Evidently, C++ will allow this, but C will not mostly because the standard isn't exact enough to allow these kinds of casts to happen safely and automatically.

MicK7:
> @MicK7, do you think C11 and _Generic are widespread enough that we can adopt it? I used _Generic to address #920. but that is still a WIP. IMHO, this is exactly what _Generic is useful for, I'm just not certain of the C11 support these days, although, HDF5 2.0 requires C11

I think that C11 is widespread enough to be adopted. For _Generic, if it is used wisely I do not see what the issue could be.

PR #930, opened 2026-01-29
Fixes for cg_conn_write to be parallel enabled

### LINKED ISSUE #929: Periodicity in 2D with Abutting1-to-1
### Discussed in https://github.com/CGNS/CGNS/discussions/928

<div type='discussions-op-text'>

<sup>Originally posted by **jrwrigh** January 27, 2026</sup>
I'm writing up a parallel reader/writer in PETSc for periodicity. If I understand the standard correctly, I need to create a `GridConnectivity_t` node (using `cg_conn_write()`), write out the `PointList` and `PointListDonor` describing the matching periodic-donor faces/edges (for 3D/2D, respectively) , then set `GridConnectivityProperty_t` with `cg_conn_periodic_write()`.

Two questions:

1. What should the `GridConnectivityType_t` be? I would think `Abutting1to1` would be appropriate (as the `PointList` and `PointListDonor` would be a 1-to-1 connectivity description), but the SIDS documentation says

> 8.4 Note 4: For `Abutting` or `Abutting1to1` interfaces, `GridLocation` can be either `Vertex` or `FaceCenter`.

For a 2D mesh, though, shouldn't the `GridLocation` be `EdgeCenter` since the boundary of the domain is made up of edges rather than faces? Or is the "face" in `FaceCenter` referring to entities one dimension lower than the `CellDim`? The documentation for `cg_boco_gridlocation_write` has 

> location – [in] (....) For CellDim=2, CellCenter refers to area elements, and the additional value EdgeCenter is allowed. (...)

which seems to mean that `EdgeCenter` would be the appropriate choice. Should the SIDS documentation read:

> 8.4 Note 4: For `Abutting` or `Abutting1to1` interfaces, `GridLocation` can be either `Vertex`, `EdgeCenter` if `CellDim = 2`, or `FaceCenter` if `CellDim = 3`.

2. I'd like to use `cgp_ptlist_write_data()` to write out the `PointList` and `PointListDonor` values. For BOCOs, we pass `NULL` to points for `cg_boco_write()` and then go back and do `cgp_ptlist_write_data()`:

```c
  if (cg_boco_write(fn, B, Z, "Left BC", CGNS_ENUMV(BCTypeUserDefined), CGNS_ENUMV(PointList), n_boco_elems, NULL, &BC))
    cgp_error_exit();
  {
    const int depth    = 4;
    char *labels[]     = {"Zone_t", "ZoneBC_t", "BC_t", "PointList"};
    int indices[] = {Z, 1, BC, 0};
    if (cg_golist(fn, B, depth, labels, indices)) cgp_error_exit();
  }
  if (cgp_ptlist_write_data(fn, start_local, end_local, el_ptr))
    cgp_error_exit();
```

But `cg_conn_write()` currently doesn't allow for `donor_data` to be `NULL`:
```c
    if (ndata_donor) {
        if (NULL == donor_data) {
            cgi_error("Invalid input: number of donor points given but data is NULL");
            return CG_ERROR;
        }
```

Should I submit a PR to remove this check and allow `cg_conn_write()` to be used similar to `cg_boco_write()`?</div>

jrwrigh:
This does several fixes in order to allow `cg_conn_write` to be used with `cgp_ptlist_*()` functions.

1. Allows for `donor_data` to be `NULL`
2. Allows for `PointList` and `PointListDonor` to be navigated to (previously could not)

Additionally, this allows for `location` to be `EdgeCenter` when `cell_dim == 2` and adds in checks for the other dimensions as well.

Fixes #929 

----

> [!IMPORTANT]
> Enhances `cg_conn_write` for parallel use by allowing `donor_data` to be `NULL`, enabling `PointList` navigation, and supporting `EdgeCenter` location for 2D in `cgnslib.c` and `cgns_internals.c`.
> 
>   - **Behavior**:
>     - `cg_conn_write` now allows `donor_data` to be `NULL` in `cgnslib.c`.
>     - Enables navigation to `PointList` and `PointListDonor` in `cgi_next_posit()` in `cgns_internals.c`.
>     - Supports `EdgeCenter` location for `cell_dim == 2` and checks other dimensions in `cg_conn_write` in `cgnslib.c`.
>   - **Misc**:
>     - Sets `conn->ptset.name` to `PointSetTypeName[ptset_type]` in `cg_conn_write` in `cgnslib.c`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=CGNS%2FCGNS&utm_source=github&utm_medium=referral)<sup> for c72cbc8853a988fe952a5cc5a04e43dbee09a983. You can [customize](https://app.ellipsis.dev/CGNS/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/src/cgns_internals.c
+++ b/src/cgns_internals.c
@@ -12546,6 +12546,27 @@ static int cgi_next_posit(char *label, int index, char *name)
                            label, 1, c->cprop->id);
             }
         }
+        else if (0 == strcmp (label, "IndexArray_t")) {
+            int current_count = 0;
+
+            /* Check PointList (Receiver) */
+            if (c->ptset.name[0] != '\0') {
+                current_count++;
+                if (index == current_count || 0 == strcmp(c->ptset.name, name)) {
+                    return cgi_add_posit(&c->ptset, label, 1, c->ptset.id);
+                }
+            }
+
+            /* Check PointListDonor (Donor) */
+            if (c->dptset.name[0] != '\0') {
+                current_count++;
+                if (index == current_count || 0 == strcmp(c->dptset.name, name)) {
+                    return cgi_add_posit(&c->dptset, label, 1, c->dptset.id);
+                }
+            }
+
+            return CG_INCORRECT_PATH;
+        }
         else if (0 == strcmp (label, "UserDefinedData_t")) {
             if (--index < 0) {
                 for (n = 0; n < c->nuser_data; n++) {
--- a/src/cgnslib.c
+++ b/src/cgnslib.c
@@ -10178,6 +10178,12 @@ int cg_conn_id(int fn, int B, int Z, int J, double *conn_id)
  *                              ptset_type of PointRange, npnts is always two. For a ptset_type of
  *                              PointList, npnts is the number of points in the PointList.
  * \param[in]  pnts             Array of points defining the interface in the current zone.
+ *                              Post 4.5.1 release, may be NULL when npnts > 0 to create the node
+ *                              structure without data (useful for parallel I/O with PointList).
+ *                              The created node name will be "PointList" (if ptset_type is PointList)
+ *                              or "PointRange" (if ptset_type is PointRange).
+ *                              When NULL, use cg_goto() to navigate to this node
+ *                              and cgp_array_write_data() to write the data in parallel.
  * \param[in]  donorname        Name of the zone interfacing with the current zone.
  * \param[in]  donor_zonetype   Type of the donor zone. The admissible types are Structured and
  *                              Unstructured.
@@ -10190,11 +10196,28 @@ int cg_conn_id(int fn, int B, int Z, int J, double *conn_id)
  *                              only for backward compatibility. The donor data is always read as cgsize_t.
  * \param[in]  ndata_donor      Number of points or cells in the current zone. These are paired with points,
  *                              cells, or fractions thereof in the donor zone.
- * \param[in]  donor_data       Array of donor points or cells corresponding to ndata_donor. Note that it is
- *                              possible that the same donor point or cell may be used multiple times.
+ * \param[in]  donor_data       Array of donor points or cells corresponding to ndata_donor.
+ *                              Post 4.5.1 release, may be NULL when ndata_donor > 0 to create the node
+ *                              structure without data (useful for parallel I/O).
+ *                              The created node name will be "PointListDonor" (if donor_ptset_type is PointListDonor)
+ *                              or "CellListDonor" (if donor_ptset_type is CellListDonor).
+ *                              When NULL, use cg_goto() to navigate to this node
+ *                              and cgp_array_write_data() to write the data in parallel.
+ *                              Note that it is possible that the same donor point or cell may be
+ *                              used multiple times.
  * \param[out] J                Interface index number, where 1 ≤ J ≤ nconns.
  * \return \ier
  *
+ * \note For parallel I/O workflows (CGNS version > 4.5.1):
+ *       1. Call cg_conn_write() with pnts=NULL and/or donor_data=NULL to create the node structure.
+ *       2. Use cg_goto() to navigate to the created "PointList" or "PointListDonor" node.
+ *       3. Call cgp_array_write_data() to write the actual index data in parallel.
+ *       This pattern follows the same approach as cg_boco_write() and other parallel I/O operations.
+ *
+ * \warning If pnts or donor_data is NULL and npnts > 0, the corresponding node will be created with
+ *          dimensions set but no data. The data MUST be written later using
+ *          cgp_array_write_data(), otherwise the file will be incomplete and reading
+ *          operations will fail or return undefined values.
  */
 int cg_conn_write(int fn, int B, int Z,  const char * connectname,
           CGNS_ENUMT(GridLocation_t) location,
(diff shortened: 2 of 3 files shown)


jrwrigh:
I'm marking this as draft for now, as I haven't full verified whether the code works, but it does at least allow my PETSc writer to execute without errors and the output CGNS file looks pretty close to correct.

src/cgns_internals.c
```diff
@@ -12546,6 +12546,14 @@ static int cgi_next_posit(char *label, int index, char *name)
                            label, 1, c->cprop->id);
             }
         }
+        else if (0 == strcmp (label, "IndexArray_t")) {
+            if ((index == 1 || 0 == strcmp (c->ptset.name, name))) {
```
jrwrigh: One thing I noticed is that the `ptset` and `dptset` in `cgns_conn` are structs. Compared to `cgns_boco`, which has it's `ptset` as pointers to struct. I don't think this is a huge issue, as they're used more or less the same way and, as far as I can tell, `cgi_add_posit()` only cares about the address to the struct. Obviously there might be some potential issues with use-after-free, but it doesn't seem like that's a risk here?

brtnfld: You must increment a counter so that when index is 2, it matches the second available node. 

Since ptset and dptset are embedded structs you need  check if they are named (logically exist), or you could check the ID.

```
else if (0 == strcmp (label, "IndexArray_t")) {
    int current_count = 0;

    /* Check PointList (Receiver) */
    if (c->ptset.name[0] != '\0') {
        current_count++;
        if (index == current_count || 0 == strcmp(c->ptset.name, name)) {
            return cgi_add_posit(&c->ptset, label, 1, c->ptset.id);
        }
    }

    /* Check PointListDonor (Donor) */
    if (c->dptset.name[0] != '\0') {
        current_count++;
        if (index == current_count || 0 == strcmp(c->dptset.name, name)) {
            return cgi_add_posit(&c->dptset, label, 1, c->dptset.id);
        }
    }

    return CG_INCORRECT_PATH;
}
```

src/cgns_internals.c
brtnfld, line 1: If you remove the NULL check, you might need to create the "PointList" node without data (empty DataArray_t) so the parallel writer can find it. For example, cgp_ptlist_write_data expects the PointList node to exist.
Check how cg_section_write handles NULL elements. Usually, it creates the node with dimensions but no data.

Now, whether our testing will pick up on this is unknown and another story... 
But I would assume it would segfault downstream.

jrwrigh: That's accounted for in my testing right now and it runs correctly (e.g. doing `cg_conn_write`, `cg_golist`, and `cgp_ptset_write_data()`). At the bottom of `cg_conn_write()`, there's:
```c
    if (npnts>0) {
        char_33 PointSetName;
        strcpy (PointSetName, PointSetTypeName[conn->ptset.type]);
        if (cgi_write_ptset(conn->id, PointSetName, &conn->ptset, index_dim,
            (void *)pnts)) return CG_ERROR;
```

which will write out the node, even if `pnts = NULL`. Same for the `PointListDonor` node. This is what `cg_boco_write()` does as well.

>  But I would assume it would segfault downstream.

It doesn't. `cgi_write_ptset()` calls `cgi_new_node()`, which will create the node and then exit with `CG_OK` if the data pointer passed to it is `NULL` before it tries to write data.

brtnfld (changes requested):
- [ ] The changes in cgns_internals.c need to be made.

- [ ] The cg_conn_write docs need to be updated:

```
 * \param[in]  pnts             Array of points defining the interface in the current zone. 
 *                              Post 4.5.1 release, may be NULL when npnts > 0 to create the node 
 *                              structure without data (useful for parallel I/O with PointList). 
 *                              The created node name will be "PointList" (if ptset_type is PointList) 
 *                              or "PointRange" (if ptset_type is PointRange).
 *                              When NULL, use cg_goto() to navigate to this node 
 *                              and cgp_array_write_data() to write the data in parallel.

* \param[in]  donor_data       Array of donor points or cells corresponding to ndata_donor. 
 *                              Post 4.5.1 release, may be NULL when ndata_donor > 0 to create the node 
 *                              structure without data (useful for parallel I/O). 
 *                              The created node name will be "PointListDonor" (if donor_ptset_type is PointListDonor)
 *                              or "CellListDonor" (if donor_ptset_type is CellListDonor).
 *                              When NULL, use cg_goto() to navigate to this node 
 *                              and cgp_array_write_data() to write the data in parallel. 
 *                              Note that it is possible that the same donor point or cell may be 
 *                              used multiple times.

*
 * \note For parallel I/O workflows (CGNS version > 4.5.1):
 *       1. Call cg_conn_write() with pnts=NULL and/or donor_data=NULL to create the node structure.
 *       2. Use cg_goto() to navigate to the created "PointList" or "PointListDonor" node.
 *       3. Call cgp_array_write_data() to write the actual index data in parallel.
 *       This pattern follows the same approach as cg_boco_write() and other parallel I/O operations.
 *
 * \warning If pnts or donor_data is NULL, the corresponding node will be created with
 *          dimensions set but no data. The data MUST be written later using 
 *          cgp_array_write_data(), otherwise the file will be incomplete and reading
 *          operations will fail or return undefined values.
```

MicK7:
Please rebase on latest develop branch

PR #935, opened 2026-02-25
Add Intel OneAPI parallel build job with HDF5 2.0.0 on Windows

### LINKED ISSUE #494: [CGNS-98] cg_open fails on Windows when given a filepath with unicode characters
> This issue has been imported from JIRA. Read the [original ticket here](https://cgnsorg.atlassian.net/browse/CGNS-98).

- _**Created at:**_ Thu, 2 Jun 2016 15:34:48 -0500

<p>We would like to request that the cg_open function be updated to allow the opening of Unicode filepaths on Windows.</p>

<p>Our product uses the CGNS library to load in CGNS data. We have recently gone through our code to update it to properly localize it to accept unicode characters, but CGNS files are still unable to be loaded when there is a unicode character in the filepath.  We have traced this into the cg_open function, which eventually calls open to load files.  For Linux and Mac this is okay, but Windows requires the filepath to be converted to wide and passed to _wopen before it will recognize the unicode characters as valid bytes in the file.</p>

### LINKED ISSUE #525: [CGNS-129] Build shared (dll) cngslib on windows
> This issue has been imported from JIRA. Read the [original ticket here](https://cgnsorg.atlassian.net/browse/CGNS-129).

- _**Created at:**_ Tue, 26 Dec 2017 17:00:38 -0600

<p>The cgns shared library (*.dll) should work out of the box.</p>

### LINKED ISSUE #582: [CGNS-187] Fully test Windows 10 for all use cases
> This issue has been imported from JIRA. Read the [original ticket here](https://cgnsorg.atlassian.net/browse/CGNS-187).

- _**Created at:**_ Wed, 13 Nov 2019 03:23:48 -0600

<p>Windows 10 is important to several commercial software vendors using CGNS.<br/>
Please test 32bit/64bit API, serial and parallel and C and Fortran APIs on Windows 10,</p>

### LINKED ISSUE #601: [CGNS-206] Windows is under-tested. Suggest Test C and Fortran serial and parallel on Windows 10 x64 as a bare minimum
> This issue has been imported from JIRA. Read the [original ticket here](https://cgnsorg.atlassian.net/browse/CGNS-206).

- _**Created at:**_ Mon, 27 Apr 2020 02:22:55 -0500

### LINKED ISSUE #608: [CGNS-213] Extend testing suite CI services
> This issue has been imported from JIRA. Read the [original ticket here](https://cgnsorg.atlassian.net/browse/CGNS-213).

- _**Created at:**_ Fri, 8 May 2020 11:40:15 -0500

<p>Expand  PR testing and develop testing to using GitHub Actions, Azure and CircleCI testing (in addition to TravisCI and AppVeyor)</p>

<p>Possibly switch from using AppVeyor to using GitHub Actions and Azuree Pipelines since they support Windows.</p>

<p>Use Azure Pipelines for linux builds as they allow 10 parallel builds.</p>

<p>MPI builds should be switched to GitHub actions. Switch to using docker images for Linux builds.</p>

### LINKED ISSUE #663: [CGNS-268] Use Intel One API in github actions to test with Fortran on Windows.
> This issue has been imported from JIRA. Read the [original ticket here](https://cgnsorg.atlassian.net/browse/CGNS-268).

- _**Created at:**_ Fri, 3 Dec 2021 17:40:50 -0600

<p>Maybe we could rely on intel OneAPI new compiler for testing. <a href="https://github.com/oneapi-src/oneapi-ci" class="external-link" rel="nofollow noreferrer">https://github.com/oneapi-src/oneapi-ci</a><br/>
It looks possible to add it to github CI : <a href="https://github.com/oneapi-src/oneapi-ci/blob/master/.github/workflows/build_all.yml" class="external-link" rel="nofollow noreferrer">https://github.com/oneapi-src/oneapi-ci/blob/master/.github/workflows/build_all.yml</a></p>

### LINKED ISSUE #809: Failing oneAPI Windows tests
Failing tests with oneAPI on Windows for develop.

The following tests FAILED:
         34 - test-pcgns_ctest (Failed)
         43 - test-test_unstruc_quad (Failed)
         48 - test-pcgns_ftest (Failed)
         52 - test-test_unstruc_quad_f90 (Failed)

test-pcgns_ctest -- passes when run individually.

48: Check cgp_coord_multi_read_data_f values                                                 *FAILED*
48:    FAILURE REPORTED --ERR: values are incorrect
48: [process 0]:no CGNS error reported
48: Abort(1) on node 0 (rank 0 in comm 0): application called MPI_Abort(MPI_COMM_WORLD, 1) - process 0
1/1 Test #48: test-pcgns_ftest .................***Failed   58.74 sec

0% tests passed, 1 tests failed out of 1

brtnfld:
Add build_windows_parallel job that builds HDF5 2.0.0 from source with Intel MPI parallel support, then builds and tests CGNS with parallel enabled using Intel compilers (icx/ifx) on Windows.

Fixes #809 , all windows tests pass (serial and parallel), added testing for Intel OneAPI on Linux
Fixes #525 , dlls are properly handled.
Fixes #601 , adds OneAPI (serial, parallel) for Windows and Linux
Fixes #582 , adds OneAPI (serial, parallel) for Windows and Linux in CI
Fixes #663 , adds OneAPI (serial, parallel) for Windows and Linux in CI
Fixes #608 , extends the testing, included complete Windows coverage in CI
Fixes #494, Added test for Unicode (UTF-8) filepath support. The test verifies creating, reading, and modifying CGNS files with non-ASCII characters in the directory and filenames, as well as external links that reference files at Unicode paths. Requires HDF5 >= 2.0.0. Linux, Mac, and Windows tested.

----

> [!IMPORTANT]
> Add Intel OneAPI parallel build job on Windows, support for Unicode file paths, and update tests and documentation.
> 
>   - **Build System**:
>     - Adds `build_windows_parallel` job in `intel.yml` for building HDF5 2.0.0 with Intel MPI on Windows.
>     - Updates URLs and versions for Intel OneAPI components in `intel.yml`.
>     - Removes obsolete scripts: `build_linux.sh`, `cache_exclude_linux.sh`, `install_linux.sh`.
>   - **Testing**:
>     - Adds `pcgns_unicode_test.c` and `test_unicode.c` to test Unicode filepath support.
>     - Updates `CMakeLists.txt` and `Makefile.in` to include new tests.
>     - Adds Fortran tests for parallel data writing using `TYPE(C_PTR)` overloads in `test_poly_unstructured_f90.F90` and `test_unstruc_quad_f90.F90`.
>   - **Documentation**:
>     - Updates `RELEASE.txt` with new features and known problems related to Unicode support.
>     - Adds `dependabot.yml` for GitHub Actions dependency updates.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=CGNS%2FCGNS&utm_source=github&utm_medium=referral)<sup> for 5252910a1f75fbc3c19c960cbfa1a3c972e7d124. You can [customize](https://app.ellipsis.dev/CGNS/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


brtnfld:
C tests all pass, Fortran, not so much ☹️

brtnfld:
It appears to be a bug in the Intel Windows Fortran optimization. If I do a -O0 for Fortran and -O2 for C, all the parallel tests pass, so most likely it is miscompiling something in the cgp_ API(s), as all the cg_ APIs had no issue with -O2.

brtnfld:
@MicK7, I think APIs like cgp_parent_data_write_f have a fundamental flaw. Passing a disassociated/null Fortran pointer to TARGET, DIMENSION(*) is undefined behaviour in Fortran-- the standard requires an actual array to be associated. The issue I was seeing with the -O2 optimization was that it exploits undefined behaviour and seems to replace it with something other than C_NULL_PTR compatible. Either we should enforce that rmin=rmax=0 results in C_NULL_PTR being set, or we should add an interface where parents is a C_PTR; not sure if we should deprecate this one for 5.0. A disassociated pointer must not be passed to any non-POINTER dummy.

```
SUBROUTINE cgp_parent_data_write_f(file_number, B, Z, S, rmin, rmax, parents, ier)
      IMPLICIT NONE
      INTEGER(CGSIZE_T), TARGET, DIMENSION(*), INTENT(IN) :: parents
```

Currently, I have it check for rmin,rmax=0, which is why the test passes with -O2. All the tests pass now.

Fortran standard rule (Fortran 2018, §15.5.2.3) is:
If the actual argument is a pointer, it shall be associated if the dummy argument is not a pointer.

Thoughts?

brtnfld:
There are 5 other cpg APIs that do the same thing. I'm thinking of creating a generic interface, where the other  "NULL" interface would not have the min, max (since not used if NULL), and the Null-argument. That would avoid breaking backward compatibility for apps that are not encountering any issues with this non-standard option, while still allowing others to be standard-compliant. The only drawback is having to use an if statement to select the correct interface. We could also add that, for the original interface, if min and max are 0, that indicates NULL, but that would still be non-standard if passing a NULL, but likely harmless. The other option is to add another generic interface that accepts a C_PTR. I think the C_PTR interface needs to be added, regardless

brtnfld:
I went with the TYPE(C_PTR) option.

brtnfld:
@jrwrigh, I think we are going to have to resolve Fortran issues with APIs like cgp_ptlist_write_data_f, as they are written such that passing a disassociated/null Fortran pointer to TARGET, DIMENSION(*) is undefined behaviour in Fortran —the standard requires an actual array to be associated. We should have indicated this with min=max=0 or some other way. It should have actually been of TYPE(C_PTR); this was an oversight. Since we are talking about the next release being a major one, we could correct this.

jrwrigh:
I have no qualms with changing the API, I don't use the Fortran interface.

MicK7:
> There are 5 other cpg APIs that do the same thing. I'm thinking of creating a generic interface, where the other "NULL" interface would not have the min, max (since not used if NULL), and the Null-argument. That would avoid breaking backward compatibility for apps that are not encountering any issues with this non-standard option, while still allowing others to be standard-compliant. The only drawback is having to use an if statement to select the correct interface. We could also add that, for the original interface, if min and max are 0, that indicates NULL, but that would still be non-standard if passing a NULL, but likely harmless. The other option is to add another generic interface that accepts a C_PTR. I think the C_PTR interface needs to be added, regardless

I did not have much time to look into the fortran API recently. But if I remind correctly it was planned to move away from C_PTR for arrays and have automatic generation of API for different supported types like what is done in HDF5. Is it still a possibility ?

brtnfld:
It is the opposite: the automatic generation in HDF5 is maintained solely for backward compatibility. All new APIs use C_PTR to avoid autogeneration, since modern Fortran now supports arrays with more than 7 dimensions, making that approach unworkable.

brtnfld:
The other option for the older APIs would not provide a FORTRAN INTERFACE, so that a null can be passed. I can bring it up at the next meeting.

brtnfld:
Added testing with Microsoft MPI on Windows, no Fortran for now.

brtnfld:
Fortran changes are scheduled for discussion at the next steering committee meeting, 4/14

MicK7:
The CI failures do not seem relevant (connection issues)

brtnfld:
@MicK7, I was thinking instead of using (_f03), which is vague, to be more explicit and use `*_cptr_data_f` for those APIs that did not have UB issues because they used c shims. So it would be,

Deprecated (_f shim kept)     	&nbsp; New preferred API
cgp_elements_write_data_f    &nbsp; cgp_elements_write_cptr_data_f
cgp_elements_read_data_f  &nbsp; cgp_elements_read_cptr_data_f
cgp_coord_write_data_f     &nbsp;   cgp_coord_write_cptr_data_f
cgp_coord_read_data_f       &nbsp;  cgp_coord_read_cptr_data_f
cgp_field_write_data_f       &nbsp;   cgp_field_write_cptr_data_f
cgp_field_read_data_f      &nbsp;    cgp_field_read_cptr_data_f

For the problematic UB ones, I updated the APIs to be PTR-only, breaking change for 5.0 for these APIs:
cgp_ptlist_write_data_f, cgp_ptlist_read_data_f, cgp_parentelements_read_data_f, cgp_poly_elements_write_data_f, cgp_parent_data_write_f 

It's not perfect, but it does not allow for UB going forward by including an interface call that still has the UB issue, while avoiding adding c shims and not being able to use modern Fortran interface features.

Thoughts? I'm done with it if you want to review it.

brtnfld:
@MicK7, any issues with my previous comment?

MicK7 (dismissed):

MicK7:
> @MicK7, any issues with my previous comment?

Sorry for the delay. Let's move forward with the fortran API.
There is no perfect solution. 
We need to clearly warn about the changes in the 5.0 release notes.

brtnfld:
@MicK7, what I don't like in the PR is introducing new cptr APIs and carrying that along forever.  And we are going to deprecate those _f versions anyway, so it will still be a pain point for users at that time.  This is a major release, so rewriting the _f functions in-place to take TYPE(C_PTR) arguments is a breaking change, but this is to be expected and justified at a major version. It also just seems confusing to have two APIs that do the same thing.

MicK7:
> @MicK7, what I don't like in the PR is introducing new cptr APIs and carrying that along forever. And we are going to deprecate those _f versions anyway, so it will still be a pain point for users at that time. This is a major release, so rewriting the _f functions in-place to take TYPE(C_PTR) arguments is a breaking change, but this is to be expected and justified at a major version. It also just seems confusing to have two APIs that do the same thing.

I agree with your point. Having two APIs is confusing. I would be in favor of a breaking change as it is a major release.

brtnfld:
@MicK7 , I've made the modifications to all the cgp_ APIs to accept TYPE(C_PTR), it is ready to go in if you want to review it again. I also added macros CGNS_VERSION_GE(maj,min) and CGNS_VERSION_LE(maj,min) so users can add them to their code to select the older APIs based on the CGNS version being used.

MicK7 (approved):

PR #938, opened 2026-03-13
fix(parallel): cgp_coord_multi_read_data should require READ, not WRITE

jrwrigh:
`cgp_coord_multi_read_data()` verifies that the file is in WRITE mode when it should verify that it's in READ mode.

Also fixed some memory leaks in the `multi` routines.


brtnfld (approved):
Good catch, thanks

PR #941, opened 2026-03-23
Fix length check validation

eirikurj:
## Description
Came across an issue migrating from CGNS 4.4 to 4.5. In my CGNS file, I have zone names that have a length of 31. When reading these names they appeared to have an incorrect length.

The issue appears to be happening in `C_F_string_chars` and `C_F_string_ptr`. If a name has exactly `len(F_string) - 1` characters (e.g., 31 chars in a `CHARACTER(len=32)` buffer), the loop exits with `i == len(F_string)`. The condition is then false, so the last character position is never written and thus retains whatever value was in memory (typically a null byte).

## Proposed fix
Updated the conditional check in both `C_F_string_ptr` and `C_F_string_chars` subroutines in `src/cgns_f.F90` to use `i<=len(F_string)` instead of `i<len(F_string)`, ensuring correct padding of Fortran strings when the string is exactly the length of the buffer. 

I did not find any tests that check this explicitly (or have a name of length 31), can propose something if needed.


brtnfld:
A test with a 31-character zone name (or base name) read back through the Fortran interface would be a valuable addition to prevent regression. Something like:

! Create a zone with a 31-character name, read it back, verify trim length
character(len=32) :: zone_name
zone_name = "1234567890123456789012345678901"  ! 31 chars
! ... write and read back, verify len_trim(zone_name) == 31

Otherwise, looks good!

eirikurj:
@brtnfld I have added a test. Not sure if this is what you had in mind, but please let me know if anything is missing or you think this should be done in a different way.

brtnfld:
I don't think it needs a separate test; it can be included in cgwrite_f03.F90. Here is the patch; it fails without a fix.

[test_name_length.patch](https://github.com/user-attachments/files/26278513/test_name_length.patch)

eirikurj:
I have reverted the test I added and applied your patch.

brtnfld (approved):

PR #944, opened 2026-03-28
fix(parallel): Test correct dimensions only if given non-null data ptrs

jrwrigh:
This mimics the behavior of `cgp_{field,coords}_{write,read}_data()`, in that if the passed pointer is NULL, then it assumes that no data will be written by the current process and thus it doesn't need to check the bounds.

### DIFF
--- a/src/pcgnslib.c
+++ b/src/pcgnslib.c
@@ -3246,12 +3246,18 @@ int cgp_coord_multi_read_data(int fn, int B, int Z, int *C, const cgsize_t *rmin
       }
     }
 
+    int has_data = 0;
+    for (n = 0; n < nsets; n++) {
+      if (buf[n]) { has_data = 1; break; }
+    }
     for (n = 0; n < zone->index_dim; n++) {
       dims[n] = zone->nijk[n] + zcoor->rind_planes[2*n] +
         zcoor->rind_planes[2*n+1];
-      if (rmin[n] > rmax[n] || rmin[n] < 1 || rmax[n] > dims[n]) {
-        cgi_error("Invalid index ranges.");
-        goto error;
+      if (has_data) {
+        if (rmin[n] > rmax[n] || rmin[n] < 1 || rmax[n] > dims[n]) {
+          cgi_error("Invalid index ranges.");
+          goto error;
+        }
       }
     }
 
@@ -3352,13 +3358,19 @@ int cgp_coord_multi_write_data(int fn, int B, int Z, int *C, const cgsize_t *rmi
       }
     }
 
+    int has_data = 0;
+    for (n = 0; n < nsets; n++) {
+      if (buf[n]) { has_data = 1; break; }
+    }
     for (n = 0; n < zone->index_dim; n++) {
         dims[n] = zone->nijk[n] + zcoor->rind_planes[2*n] +
                                   zcoor->rind_planes[2*n+1];
+      if (has_data) {
         if (rmin[n] > rmax[n] || rmin[n] < 1 || rmax[n] > dims[n]) {
-            cgi_error("Invalid index ranges.");
-            goto error;
+          cgi_error("Invalid index ranges.");
+          goto error;
         }
+      }
     }
 
     for (n = 0; n < nsets; n++) {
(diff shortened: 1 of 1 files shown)


src/pcgnslib.c
```diff
@@ -3461,11 +3455,13 @@ int cgp_field_multi_write_data(int fn, int B, int Z, int S, int *F,
 
       /* verify that range requested does not exceed range stored */
       for (m = 0; m < field->data_dim; m++) {
-        if (rmin[m] > rmax[m] ||
+        if (buf[m]) {
```
brtnfld: The m (dimension) index is used to check whether the m-th buffer is non-NULL, but buf is indexed by set. The check should be if (buf[n]), skip the range validation for set n if its buffer is NULL.

src/pcgnslib.c
```diff
@@ -3562,11 +3558,13 @@ int cgp_field_multi_read_data(int fn, int B, int Z, int S, int *F,
 
     /* verify that range requested does not exceed range stored */
     for (m = 0; m < field->data_dim; m++) {
-      if (rmin[m] > rmax[m] ||
-	  rmax[m] > field->dim_vals[m] ||
-	  rmin[m] < 1) {
-	cgi_error("Invalid range of data requested");
-	goto error;
+      if (buf[m]) {
```
brtnfld: same, this gets missed by the test because nsets=3 for X/Y/Z and zone->index_dim=3 for a 3D zone, and these happen to be equal

src/pcgnslib.c
```diff
@@ -3249,9 +3249,11 @@ int cgp_coord_multi_read_data(int fn, int B, int Z, int *C, const cgsize_t *rmin
     for (n = 0; n < zone->index_dim; n++) {
       dims[n] = zone->nijk[n] + zcoor->rind_planes[2*n] +
         zcoor->rind_planes[2*n+1];
-      if (rmin[n] > rmax[n] || rmin[n] < 1 || rmax[n] > dims[n]) {
-        cgi_error("Invalid index ranges.");
-        goto error;
+      if (buf[n]) {
```
brtnfld: The fix needs to introduce a separate boolean check before the dimension loop, using the set index (not the dimension index) to scan buf.

```
int has_data = 0;
for (n = 0; n < nsets; n++) {
  if (buf[n]) { has_data = 1; break; }
}
for (n = 0; n < zone->index_dim; n++) {
  dims[n] = zone->nijk[n] + zcoor->rind_planes[2*n] +
    zcoor->rind_planes[2*n+1];
  if (has_data) {
    if (rmin[n] > rmax[n] || rmin[n] < 1 || rmax[n] > dims[n]) {
      cgi_error("Invalid index ranges.");
      goto error;
    }
  }
}
```

brtnfld: same pattern applies to cgp_coord_multi_write_data

brtnfld (changes requested):

brtnfld:
Thanks for the fix — the intent is correct. However, there are a couple of issues with the current implementation:

### 1. Coord functions: missing braces on `has_data` loop

```c
if (buf[n]) has_data = true; break;
```

The `break` is outside the `if`, so the loop always exits after checking only `buf[0]`. Needs braces:
```c
if (buf[n]) { has_data = true; break; }
```

### 2. Coord functions: `bool` not available

`pcgnslib.c` doesn't include `<stdbool.h>` and doesn't use `bool`/`true`/`false` anywhere else. This will fail to compile depending on the C standard. Use `int` with `0`/`1` instead, consistent with the rest of the file.

---

Here is a patch (against `develop`) that addresses the above issues and also adds a test to `pcgns_ctest.c` that exercises NULL buffers in the multi functions. The test has rank 0 write real data while rank >= 1 passes all-NULL buffers, then reads back and verifies. This would have caught the original bug.

[fix-multi-null-buf-check.patch.gz](https://github.com/user-attachments/files/26388646/fix-multi-null-buf-check.patch.gz)

jrwrigh:
Bleh, that's embarrassing. My bad on the bugs.

brtnfld (approved):

PR #945, opened 2026-04-01
fix(parallel): Handle NULL buffers in multi-dataset collective I/O

brtnfld:
## Summary

Fixes a deadlock (and latent segfault) when ranks pass NULL data buffers to the multi-dataset parallel I/O functions, and adds a dedicated test suite covering those scenarios.

### Problem

Calling any of the `cgp_{coord,field,array}_multi_{write,read}_data` functions with NULL buffer pointers caused two distinct bugs:

1. **MPI deadlock** – the collective HDF5 I/O path created non-empty hyperslab selections for all ranks, including ranks with no data. Ranks that had nothing to write/read still entered `H5Dwrite_multi`/`H5Dread_multi` with a live selection, causing the collective operation to hang.
2. **Latent segfault** – the previous `has_data` fix used a single rank-wide flag. If a caller passed a mix of valid and NULL buffers in one call (e.g., `buf[0]=valid, buf[1]=NULL`), `has_data` was `1` and `H5Sselect_hyperslab` was applied to *all* datasets including the NULL one. HDF5 would then dereference the NULL pointer.

### Changes

| File | Change |
|------|--------|
| `src/pcgnslib.c` | Replace global `has_data` flag with per-dataset `has_data_k` check inside `readwrite_multi_data_parallel`. Apply `H5Sselect_none` on both mem and file spaces for each dataset whose buffer is NULL. Also adds NULL-buffer guards to `cgp_array_multi_{write,read}_data` and `cgp_particle_{coord,field}_multi_{write,read}_data`, which were missed by PR #944. |
| `src/ptests/test_multisets.c` | New dedicated test file covering: (1) all ranks contributing data (round-trip correctness), (2) mixed NULL – all ranks except the last contribute; last rank passes NULL, (3) rank-0-only – only rank 0 contributes; all others pass NULL. Tests coords, fields, and arrays in each scenario. |
| `src/ptests/pcgns_ctest.c` | Remove the `multisets()` function and its NULL-buffer variants (now in `test_multisets.c`). |
| `src/ptests/Makefile.in` | Replace 14 identical single-source C build stanzas with one `%$(EXE): %.c` pattern rule. |
| `src/ptests/CMakeLists.txt` | Register `test_multisets` and increase parallel rank count from 2 → 4 to match autotools CI. |

### Test coverage

`test_multisets` runs under `mpiexec -n 4` and exercises:
- `cgp_coord_multi_{write,read}_data`
- `cgp_field_multi_{write,read}_data`
- `cgp_array_multi_{write,read}_data`

for three NULL-buffer scenarios. All data is verified on read-back.

### Related
Fixes the NULL-buffer gap left by #944.


brtnfld:
@MicK7, I think this now covers and tests various edge cases.

MicK7 (commented):
LGTM

MicK7 (approved):

PR #953, opened 2026-05-10
fix: resolve CodeQL format specifier and overflow warnings in cgnslib.c

brtnfld:
Replace %ld format specifiers for cgsize_t arguments with the project's PRIdCGSIZE macro, which expands correctly for both 32-bit ("d") and 64-bit (PRId64) builds. Also fix a potential integer overflow in the ParentElements malloc call by casting cnt and newsize to size_t before multiplication.


brtnfld:
@MicK7, do those last three PRs look ok? I will next work on resolving #935

MicK7 (approved):

